Issue #54, opened 2016-01-04
mm.audio.ffmpeg.get_file_info fails extracting sample_rate when using avprobe


fdlm:
avprobe (version 9.18) prints sample_rate as a float:
```
[streams.stream.0]
index=0
codec_name=flac
codec_long_name=FLAC (Free Lossless Audio Codec)
codec_type=audio
codec_time_base=1/44100
codec_tag_string=[0][0][0][0]
codec_tag=0x0000
sample_rate=44100.000000
channels=1
bits_per_sample=0
avg_frame_rate=0/0
time_base=1/44100
start_time=N/A
duration=216.685714
```
which is why
```python
info['sample_rate'] = int(line[len('sample_rate='):])
```
does not work.

PR #55, opened 2016-01-04
get_file_info now works with avprobe

fdlm:
Solves issue #54.

Made the cast explicitly only if the function is using `avprobe` so we can remove it if we decide to drop support for `avprobe`.

### DIFF
--- a/madmom/audio/ffmpeg.py
+++ b/madmom/audio/ffmpeg.py
@@ -295,7 +295,10 @@ def get_file_info(infile, cmd='ffprobe'):
         if line.startswith(b'channels='):
             info['num_channels'] = int(line[len('channels='):])
         if line.startswith(b'sample_rate='):
-            info['sample_rate'] = int(line[len('sample_rate='):])
+            # the int(float(...)) conversion is necessary because
+            # avprobe returns sample_rate as floating point number
+            # which int() can't handle.
+            info['sample_rate'] = int(float(line[len('sample_rate='):]))
     # return the dictionary
     return info
 


madmom/audio/ffmpeg.py
```diff
@@ -295,7 +295,10 @@ def get_file_info(infile, cmd='ffprobe'):
         if line.startswith(b'channels='):
             info['num_channels'] = int(line[len('channels='):])
         if line.startswith(b'sample_rate='):
-            info['sample_rate'] = int(line[len('sample_rate='):])
+            if cmd == 'avprobe':
+                info['sample_rate'] = int(float(line[len('sample_rate='):]))
```
superbock: Does a single line `int(float(line[len('sample_rate='):]))` also work for all cases?
If yes, we could skip the if/else.

I'd like to have a test for this, but it is rather difficult. Because of the other avconv/avprobe issues, TravisCI uses the "real" ffmpeg, otherwise other tests fail.

fdlm: Yes, the single line would probably also work, although I can't test it with ffmpeg.
If we convert it to one line, we should add a comment explaining why the `int(float(...))` thingy is necessary (because of avprobe)

superbock: +1 for adding an explicit comment. If it works for avprobe, just go ahead and change it, ffmpeg gets tested automatically.

superbock: Verified that it works with ffmpeg, too.

fdlm:
I think this can be merged now

superbock:
fixes #54, thanks!

PR #76, opened 2016-02-03
Added downbeat switch to beat evaluation

flokadillo:
If downbeat flag is true, downbeats instead of beats are read from file.

### DIFF
--- a/madmom/evaluation/beats.py
+++ b/madmom/evaluation/beats.py
@@ -55,7 +55,7 @@ class BeatIntervalError(Exception):
 
 
 @suppress_warnings
-def load_beats(values):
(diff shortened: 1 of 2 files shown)


superbock:
Instead of using the "update branch" button it's better to rebase the whole branch on top of the develop branch. To do so, please follow these steps:

```sh
git checkout develop
git pull
git checkout feature/downbeat_eval
git rebase --onto NEWBASE OLDBASE
```

with ``NEWBASE`` being the latest commit on develop and ``OLDBASE`` the parent of your first commit in this branch. I'm not sure whether ``develop`` can be used as ``NEWBASE``, probably yes.

PR #95, opened 2016-02-19
Remove the option to normalize the observations of the ObservationModels

superbock:
Needs to be confirmed that it's not needed for the `PatternTrackingProcessor` any more.


fdlm:
Just took a quick look at the diff. I also think the `norm_observations` parameter should be removed from the `GMMPatternTrackingObservationModel`. Normalising the inputs to an HMM is part of preprocessing the data, rather than the HMM (and the observation model is part thereof) itself. The `PatternTrackingProcessor` should do that itself if it is necessary.

Issue #99, opened 2016-02-29
Rework the .hdf5/.npz model format


superbock:
Right now the model format is suited only for RNNs (for historical reasons), but we should extend/rework it in a way it is able to do everything. I hereby propose the following:

* we keep .hdf5/.npz as our primary formats, since this works quite well
* to flatten a .hdf5 file to .npz (a flat format) we use the names (HDF5_object.name) as keys
* objects and functions are stored as a group (arbitrary names are allowed)
* the arguments needed to instantiate an object are stored as datasets with the name of the argument
* in case a list of objects is needed as an argument, these get nested in a group again (with the argument as the group name)
* to determine which class/type an object/function is of we save it as "type" attribute
* HDF5 attributes are stored with key "attrs/attribute_name" in the NPZ dictionary

Questions:
* how do we store the order of the items in a lists? Should we attach an "id" attribute to the group or rather encode the order in the group name (making it sortable)?
* how do we determine if an instance of the class is wanted or just the class/function itself?

Example:

```
h5py.File
> Group: rnn
  > Attribute: type='madmom.ml.rnn.RecurrentNeuralNetwork'
  > Group: layers
    > Group: hidden_layer_0
      > Attribute: type='madmom.ml.rnn.BidirectionalLayer'
      > Attribute: id=0
      > Group: fwd_layer
        > Attribute: type='madmom.ml.rnn.RecurrentLayer'
        > Dataset: weights
        > Dataset: bias
        > Dataset: recurrent_weights
        > Group: transfer_fn
          > Attribute: type='madmom.ml.rnn.tanh'
      > Group: bwd_layer
        > Attribute: type='madmom.ml.rnn.RecurrentLayer'
        > Dataset: weights
        > Dataset: bias
        > Dataset: recurrent_weights
        > Group: transfer_fn
          > Attribute: type='madmom.ml.rnn.tanh'
    > Group: hidden_layer_1
      > Attribute: type='madmom.ml.rnn.BidirectionalLayer'
      > Attribute: id=1
      > Group: fwd_layer
        > Attribute: type='madmom.ml.rnn.RecurrentLayer'
        > Dataset: weights
        > Dataset: bias
        > Dataset: recurrent_weights
        > Group: transfer_fn
          > Attribute: type='madmom.ml.rnn.tanh'
      > Group: bwd_layer
        > Attribute: type='madmom.ml.rnn.RecurrentLayer'
        > Dataset: weights
        > Dataset: bias
[...]

PR #101, opened 2016-03-05
Refactor NN and model io

fdlm:
This is an attempt to refactor the neural network stuff and model input/output (see issue #99). The feature is **not ready yet**, I just want a place to discuss stuff about how to proceed.

**TODO** (feel free to add more if you find something)

- [ ] Write a function that converts old models to the new format
- [ ] Implement model loading from `npz` files
- [ ] Rewrite the comments and documentation of the NN part (they do not make sense anymore)
- [ ] Adapt scripts in `bin/` directory

**Questions/Remarks**

 - I made an initial attempt at restructuring the code. I like the way Lasagne does this, with submodules like `layers`, `nonlinearities`, etc. I think it is nice to have the Network classes like `RecurrentNeuralNetwork` directly in the `madmom.ml.nn` namespace, thus the import in `__init__.py`
 - What do you think about changing the name 'transfer function' to 'nonlinearity', the way they are called in Lasagne?


superbock:
Let me just add a few things that come to my mind:

1) I like the structure with `ml.nn.nets` and `ml.nn.layers`
2) I don't like `nonlinearities`
3) I'd collect the activation functions in a `ml.nn.activations` module, simply because everybody calls them activation functions (see https://en.wikipedia.org/wiki/Activation_function)
4) I would rather see #99 (the HDF5/NPZ stuff) decoupled from the `ml.nn` refactoring and leave the functionality for this PR as it is, just refactor everything and add tests
5) If it is easier, leave the `.pxd` stuff out for now, we can add it later on. We should test where this is needed, my first guess is that it only helps with the `for` loops.
6) We can then rename the `transfer_fn` (together with #99), but I'd rename it to `activation_fn` or `activation_function` (see 3) above)
7) I don't like to have the reordering of the imports in `/bin` in this PR

fdlm:
You are right, the new model structure should go into a separate PR, see #102.

The remaining stuff:

 - OK, let's call the sub-module `activations` and the parameter `activation_fn`.
 - PyCharm did the reordering and I did not notice. It should stay the way it was, I agree!

PR #102, opened 2016-03-06
Load functions for new model format

fdlm:
Attempts to address #99 

Model loading should work fine.

**TODO**

- [ ] Rewrite docstring of `madmom/ml/io.py`. It should contain a description of the new model format (Does the description in #99 suffice?)
- [ ] Write a function that converts old model files to the new format. *Alternative:* Write a `dump` function that writes an existing model in the new format. The conversion function could then just use the existing functionality of the RNN's `load` method to load the model, and dump it into the new format.
- [ ] Remove `load` method in `RecurrentNeuralNetwork`
- [ ] Adapt all scripts that load models


superbock:
Thanks for doing this separately! Again a few comments from my side:

1. I'm not sure if we need a function to convert the old model files to the new format. I guess that I'm the only one so far who stores models. I will update my code to export the models in the new format and that's it. For the future, we must offer a conversion function only if the format changes (which should be not the case if we do it properly :smirk: )
2. `h5py` offers a dictionary like way to access the groups and items, thus much of the code for loading models could be unified - no matter if they are stored as .hdf5 or .npz
3. There shouldn't be a need to modify the scripts in bin, at least I can't see it right now
4. Should we move the loading functionality to the `Processor` class? This would be the most universal approach. So any `Processor` could be loaded universally, no matter how it is stored (pickled, as .hdf5, or .npz). The only thing we would need to change ist that a layer inherits from Processor since any `Layer` _is_ basically a `Processor`.

PR #129, opened 2016-04-12
Feature/deep chroma

fdlm:
This adds a Chroma extractor based on a Deep Neural Network.


fdlm:
made obsolete by #148

PR #132, opened 2016-04-28
Copy downbeat tracking models on install

ains:


superbock:
Thanks for the PR, I indeed missed that. But I'm not sure if this is the correct solution anyways. PR #125 changed the way the models are handled, so we have to include `__init__.py` as well. I think the correct (and easiest) solution would be to simply include all files in the `models` directory.

superbock:
Just checked, the models' `__init__.py` is included anyways, so this solution works. Thanks!

PR #152, opened 2016-07-07
Chord Recognition with CNN and CRF

fdlm:
This is the finalised (?) cnn chord recognition functionality.

Also, addresses #153.


tests/test_features_chords.py
```diff
@@ -0,0 +1,104 @@
...
+
+from madmom.features import Activations
+from madmom.features.chords import *
+
+
+sample_file = "%s/sample.wav" % AUDIO_PATH
+sample_cnn_act = Activations('%s/sample.cnn_chord_features.npz' %
+                             ACTIVATIONS_PATH)
+sample_deep_chroma_act = Activations('%s/sample.deep_chroma.npz' %
+                                     ACTIVATIONS_PATH)
```
superbock: Please use `os.path.join()` as in `tests/test_utils.py` so Windows tests do not fail because of this -- they do for some other reason nonetheless...

fdlm: Good point. I'll also update `test_features_beats.py` then...

bin/CNNChordRecognition
```diff
@@ -0,0 +1,91 @@
...
+    extracted chords to same location as the audio files.
+
+    The 'pickle' mode can be used to store the used parameters to be able to
+    exactly reproduce experiments.
+
+    '''
+    )
+    # version
+    p.add_argument('--version', action='version',
+                   version='CNNChordRecogniser.2016')
```
superbock: Should be 'CNNChordRecognition'

madmom/utils/__init__.py
```diff
@@ -405,6 +405,37 @@ def quantize_events(events, fps, length=None, shift=None):
...
+    -----
+    Segment files cannot contain comments, because e.g. chord annotations
+    can contain the '#' character! The maximum label length is 64 characters.
+
+    """
+    return np.loadtxt(filename, comments='', dtype=[('start', np.float),
+                                                    ('end', np.float),
+                                                    ('label', 'S64')])
+
+
```
superbock: Maybe this function should be moved to `features.chords` since it is quite specific to chords.

PR #155, opened 2016-07-08
unify sep & delimiter argument of writer functions

superbock:


superbock:
@fdlm: this changes the output of the `DCChordRecognition` program to use a tab instead of space to separate the values. Don't know if this is ok for you. Tabs are the default for all other programs.

superbock:
Ups, wanted to merge #156, not this one. We have to address the tab/space output format of `DCChordRecognition` in a separate PR if this is an issue.

Issue #145, opened 2016-07-01
pip install madmom fails on Windows


steiml:
Hi,

I tried to install madmom as described in the user guide but it fails with a compiler error:
C:\Users\michael.stein>pip install madmom
 madmom/features\beats_crf.c(3832) : error C2065: 'INFINITY' : undeclared identifier
madmom/features\beats_crf.c(3967) : error C2065: 'INFINITY' : undeclared identifier

Full output is pasted below

Best,
Michael

Microsoft Windows [Version 6.1.7601]
Copyright (c) 2009 Microsoft Corporation.  All rights reserved.

C:\Users\michael.stein>pip install madmom
Collecting madmom
  Using cached madmom-0.13.2.tar.gz
Building wheels for collected packages: madmom
  Running setup.py bdist_wheel for madmom ... error
  Complete output from command c:\users\michael.stein\anaconda2\python.exe -u -c "import setuptools, tokenize;__file__='
c:\\users\\michae~1.ste\\appdata\\local\\temp\\pip-build-dijxcx\\madmom\\setup.py';exec(compile(getattr(tokenize, 'open'
, open)(__file__).read().replace('\r\n', '\n'), __file__, 'exec'))" bdist_wheel -d c:\users\michae~1.ste\appdata\local\t
emp\tmp8ropqmpip-wheel- --python-tag cp27:
  running bdist_wheel
  running build
  running build_py
  creating build
  creating build\lib.win-amd64-2.7
  creating build\lib.win-amd64-2.7\madmom
  copying madmom\processors.py -> build\lib.win-amd64-2.7\madmom
  copying madmom\__init__.py -> build\lib.win-amd64-2.7\madmom
  creating build\lib.win-amd64-2.7\madmom\audio
  copying madmom\audio\ffmpeg.py -> build\lib.win-amd64-2.7\madmom\audio
  copying madmom\audio\filters.py -> build\lib.win-amd64-2.7\madmom\audio
  copying madmom\audio\signal.py -> build\lib.win-amd64-2.7\madmom\audio
  copying madmom\audio\spectrogram.py -> build\lib.win-amd64-2.7\madmom\audio
  copying madmom\audio\stft.py -> build\lib.win-amd64-2.7\madmom\audio
  copying madmom\audio\__init__.py -> build\lib.win-amd64-2.7\madmom\audio
  creating build\lib.win-amd64-2.7\madmom\evaluation
[...]

PR #157, opened 2016-07-08
Fix windows compile errors

superbock:
Commit 0fc70e2 may bring back the old error fixed in #88, but despite all efforts we are not able reproduce this error. Hoping the best and that it may not break things.


superbock:
I will merge this branch, although it does not fully resolve #145. At least it is a step towards Windows compatibility...

PR #181, opened 2016-07-21
Revert "Fixed deep chroma model and adapted respective test data"

superbock:
Reverts CPJKU/madmom#179, since it references a non-existing submodule commit. Weird that Travis was happy with it...


superbock:
But Travis is unhappy about reverting it... closing. Fixing it with new PR.

Issue #153, opened 2016-07-08
missing program tests


superbock:
At least the `DCChordRecognition` misses tests in `tests/test_bin.py`. Maybe some others as well...

PR #183, opened 2016-07-21
Test for DCChordRecognition program

fdlm:
Also, added second audio sample file with harmonic content.

Addresses #153.


fdlm:
This is now part of #152. Closing.

Issue #33, opened 2015-12-03
add **kwargs to process()


superbock:
Add/pass `**kwargs` to/from the `process()` methods of all processors. This is needed if we want to be able to set/change/overwrite some processing options during run time.

superbock:
An easy solution would be to build a dict from the saved Processor attributes, update it with kwargs and pass it to the data class, e.g.:

```
def process(self, data, **kwargs)
    _kwargs = dict(attr1=self.attr1, attr2=self.attr2)
    _kwargs.update(kwargs)
    return DataClass(data, **_kwargs)
```

superbock:
During working on streaming mode (#185) I wanted to address this issue and incorporating the proposal above.

Unfortunately, only the behaviour of the current processor can be influenced, since passing of ``kwargs`` to the next processor does not work. This is due to the fact that functions can also act as processors and they don't accept arbitrary additional arguments.

Issue #122, opened 2016-03-26
add streaming mode


superbock:
In the long term, we should implement a "streaming mode". I.e. `Processors` should be able to process the data in chunks / blocks of arbitrary size. I'm thinking about a `block_size` parameter which can be either set to a certain size (e.g. 1 to process every frame individually) or `None` to indicate processing of the whole data at once (needed for bidirectional networks or Viterbi decoding). Setting the default to `None` would not change the current behaviour at all, but using smaller sizes could save a lot of memory.

PR #185, opened 2016-07-21
Streaming mode

superbock:
This is a stripped down version of @amaurydurand's work (PR #184).

TODO:
- [x] move buffer initialisation to subclassed of `BufferProcessor`
- [x] reduce CPU load (re-use `Filterbank`)
- [x] rewrite `bin_frequencies` as an property
- [x] use `float32` instead of `int16` dtype
- [x] ~~rewrite all features as processors (mostly `madmom.features.onsets.*`)~~ (separate issue/PR)
- [x] init `Stream` with the given arguments ~~if one is given check if the arguments match the ones of the processor~~
- [x] float `hop_size` not possible, raise error? (see issue #249)
- [x] difficult to get `diff_ratio` right in `SpectrogramDifference`, find a solution (maybe only allow `diff_frames` in online mode?)
- [x] check memory consumption
- [x] ~~make `block_size` configurable (e.g. it could be advantageous to process everything in blocks of 2048 frames instead of whole length pieces)~~ (see separate issue #250)
- [x] fix output writers to work with continuous/steady data (flush more often?)
- [x] recheck cached attributes/variables if they change
- [x] add option to reset processors (e.g. for processing of next song)
- [x] ~~rewrite `FilterbankProcessor` to process the input and create a `Filterbank` when called the first time and cache it for consecutive calls (like a buffer)~~ (see separate issue #248)
- [x] ~~drop the combined spectrogram processors (`FilteredSpectrogramProcessor`, `LogarithmicSpectrogramProcessor` and `LogarithmicFilteredSpectrogramProcessor`) altogether?~~ (see separate issue #248)
- [x] ~~add support to select mono/stereo/left/right channel in `Stream`~~ (should be handled by `Signal`, see separate issue #211)
- [x] check if passing `**kwargs` to `process()` methods is needed (related: #33)


winchell:
Will this make the existing algorithms work with RT audio input?

superbock:
Yes, exactly. We hope to get this done by end of the year.

superbock:
Rebased everything on top of #244, because this PR is basically ready -- but these two PRs could be merged in any order basically.

Besides the TODOs above, there are still a couple of things which need to be addressed/discussed:

- Do we want to have an `online` parameter in all classes to decide the mode they should be processed? For some classes (e.g. `RNNOnsetProcessor`) this would be advantageous, since they need to load different models or set some parameters accordingly.
- On the other hand, some other classes (e.g. `SpectrogramDifferenceProcessor`) have the same parameters in online/offline mode, but behave differently.

Right now `self.online` is used to call either `process_sequence()` or `process_step()` from inside the `process()` method. Theoretically, we could decouple the two things, by keeping the `online` argument around and call `process_step()` directly from `madmom.processors.process_online()`. Any thoughts?

superbock:
After thinking about it a while I think that we should split the two things (i.e. online-mode and stepwise processing).

I see the following advantages:

- we can process online processing chains not only frame-by-frame but also as a whole, which saves a lot of time, especially in batch mode
- we have to add `online` parameters only to those classes which show different behaviour in online/offline mode (e.g. different parameters)
- by defining a `process_step()` method to process everything stepwise we can explicitly define how classes should behave
- if the stepwise processing is the same as processing the whole sequence at once, we can simply set an alias `process_step = process`
- classes not having this alias are not suited/ready for stepwise processing
- the only thing we have to change are the `SequentialProcessor` and `ParallelProcessor` to have a `process_step()` method which calls `process_step()` of the individual processors

As a bonus we could change the `online` processing mode of programs (e.g. `DBNBeatTracker online`) to process everything in real-time, even when called with an input file -- if technically possible.

superbock:
Bummer! We can't do it this way, at least not if we want to be able to use functions as processors -- which is a must-have.

So we have to figure out a way which only calls a processor -- maybe even without passing any additional arguments since a lot of functions can't deal with it.

Maybe we should reconsider the original idea of #122 to use a `block_size`, although this doesn't work with normal functions either. But at least we can then use partials, to be able to call these functions...

Any thoughts on how to elegantly solve this issue?

superbock:
Another possibility is to infer the mode directly from the shape of the data. If only a single frame/value is given we process everything stepwise (i.e. a single step), if a sequence is given (i.e. one additional dimension), we process the sequence.

This could work also if we use the `online` argument solely to indicate how certain parameters should be set, and not to decide if we process the step by step or everything at once.

superbock:
Besides the open TODOs below, this PR is more or less ready.

As decided/merged in #244, `process()` is called the exact same way as before, independently whether a sequence gets processed as a whole or frame-by-frame. Stateful processors accept an additional `reset` argument, which defaults to `True` and thus does not change existing behaviour.

As discussed in #246, an additional `online` processing mode is added which processes live microphone input or files always on a frame by frame basis. The existing `single` now has the additional `--online` argument to switch between `offline` (default) and `online` parameters. Since `single` mode always loads the processors, there's no need to reset during operation. On the other hand, this enables us to allow block-wise processing later on (#250).

Open questions / TODOs:
- [ ] add option to define length of the `Stream` to retrieve only a certain number of frames (or specify a length, maybe via `stop`)
- [ ] some processors can not work online, raise error?
- [ ] how to signal buffer under-/overflow?
- [ ] how to terminate a `Stream`?

Maybe these can remain open or don't need to be addressed (right now), as this PR handles only the common foundation for various online processing methods (e.g. #238 for online beat tracking).

superbock:
One more thing to consider: at the moment there is no way to determine whether a processor is online-capable or not.

A previous attempt to solve this used different process methods. E.g. `process_sequence()` and `process_step()` for offline/online mode, respectively. However, since normal functions must be usable as processors also, Processors are simply called (like functions), thus there is not really a uniform way to distinguish between online/offline mode for processors.

There's no need to address this within this PR, since none of the added functionality affects this. We should just keep this in mind.

madmom/audio/signal.py
```diff
@@ -1364,21 +1374,22 @@ def process(self, data, **kwargs):
...
         -------
         frames : :class:`FramedSignal` instance
             FramedSignal instance
 
         """
+        # update arguments passed to FramedSignal
+        args = dict(frame_size=self.frame_size, hop_size=self.hop_size,
+                    fps=self.fps, origin=self.origin, end=self.end,
+                    num_frames=self.num_frames)
+        args.update(kwargs)
```
fdlm: Is there a reason for this change? I actually think the previous code was more direct and thus easier understandable.

madmom/audio/spectrogram.py
```diff
@@ -611,18 +613,31 @@ def __new__(cls, spectrogram, log=LOG, mul=MUL, add=ADD, **kwargs):
...
+
+    @property
+    def filterbank(self):
+        """Filterbank."""
+        return self.spectrogram.filterbank
+
+    @property
+    def bin_frequencies(self):
+        """Bin frequencies."""
+        try:
```
fdlm: Is this necessary? As far as I understand, the `self.spectrogram` is a `FilteredSpectrogram` (and thus has a filterbank), then `self.spectrogram.bin_frequencies` will return the `center_frequencies` of the filterbank anyways.

superbock: You are right. I think you meant: "*if* `self.spectrogram` is a `FilteredSpectrogram`", though.

Issue #220, opened 2016-10-06
Check for file handles in audio.signal.load_audio_file


flokadillo:
`audio.signal.load_audio_file(filename, ...)` checks if the input is a filehandle using `if not isinstance(filename, str)`. However, if _filename_ is represented as type _unicode_, this condition is true as well and we therefore conclude that the variable is a file handle and try to close the file, which results in an error.
Therefore, I suggest it would be safer to check for file handles using
`if hasattr(filename, 'read')` or `if isinstance(filename, file)`.

superbock:
You have to be careful, Python 3 does not have a `unicode` type, just `str`. Maybe just use a `try/except`.

PR #223, opened 2016-10-25
fixes error loading unicode filenames

flokadillo:
This addresses issue #220. Closing the file is now in a try block.

### DIFF
--- a/tests/test_audio_signal.py
+++ b/tests/test_audio_signal.py
@@ -11,6 +11,7 @@ import os
 import tempfile
 import unittest
 from os.path import join as pj
+import sys
 
 from . import AUDIO_PATH, DATA_PATH
 from .test_audio_comb_filters import sig_1d, sig_2d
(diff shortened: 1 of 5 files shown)


tests/test_audio_signal.py
```diff
@@ -723,6 +724,9 @@ def test_types(self):
         self.assertIsInstance(signal, np.ndarray)
         self.assertTrue(signal.dtype == np.int16)
         self.assertTrue(type(sample_rate) == int)
+        if sys.version_info[0] == 2:
+            # test unicode string type (Python 2 only)
+            signal, sample_rate = load_audio_file(unicode(sample_file))
```
superbock: Is there a reason to check this only for Python 2? Shouldn't this be checked for Python 2, too?

flokadillo: Python 3 has no `unicode` type, therefore I cannot execute `unicode(sample_file)`.

fdlm: In Python 3, all strings are unicode by default.

PR #227, opened 2016-10-31
use new Cython build system; fixes #225

### LINKED ISSUE #225: Cython 0.25 breaks build
When building madmom (or installing via pip) with Cython==0.25, it won't build and produce the following message : 

`creating build/lib.macosx-10.12-x86_64-3.5/madmom/ml
clang -bundle -undefined dynamic_lookup -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk build/temp.macosx-10.12-x86_64-3.5/madmom/ml/hmm.o -o build/lib.macosx-10.12-x86_64-3.5/madmom/ml/hmm.cpython-35m-darwin.so
building 'madmom.ml.nn.layers' extension
error: unknown file type '.pxd' (from 'madmom/ml/nn/layers.pxd')`

The same error is generated when building on Linux

superbock:


superbock:
The build for Python 3.3 fails (https://travis-ci.org/CPJKU/madmom/jobs/171941299). Seems like the old Cython 0.22.1 shipped with Miniconda 3.3 misses `Cython.Build.build_ext`. I'm not sure whether we should add a workaround for this or abandon Python 3.3 completely. The last update (a security fix) was back in 2014. It's EOL will be reached Sept. 2017 (https://www.python.org/dev/peps/pep-0398/).

Issue #230, opened 2016-12-02
Allow for initial hidden state in RNNs


fdlm:
Madmom's RNNs do not support a learned initial hidden state. Since some deep learning frameworks (e.g. Lasagne support this, it might be useful to have such a thing.

Should we adapt the current RecurrentLayer or create a new RecurrentLayer class?

superbock:
Yes, this will be supported soon, since streaming mode (#185) requires the hidden state to be a buffer anyways.

It should be save to add a ``init`` variable (with default=None) to the existing class.

fdlm:
1. In the `GRULayer` class, this parameter is called `hid_init`, so maybe we should go with this.
1. If we want to set a default value for this parameter, it must come *after* the `activation_fn` parameter:

    ```python
    def __init__(self, weights, bias, recurrent_weights, activation_fn, hid_init=None):
    ```
    Isn't this weird? All other layers have `activation_fn` as last parameter (if they have an activation function).

superbock:
Yes, this is a bit unfortunate, but I don't see a real problem. I think it is more important that the parameters have the same ordering and naming, rather than having a certain parameter name always last.

I don't like the ``hid_init`` name anyways, but that's the way it is now. But if we want to change it, I'd propose to change it now and update all other classes consistently. This could be done together with streaming mode #185, as it requires access to previous outputs / states anyways. We can keep/access the previous output of the layers as attribute of the class. These can be initialised accordingly then.

superbock:
After digging into this issue a bit I realise that we will move straight towards/into if/elif/else-hell if we implement this right.

The problem is the following:
- if we want to save the state, we need to add another attribute, which is not available in the saved models
- if we want to initialise these states, we need to save this information independently (otherwise we can not reset to this initial state later on)
- both attributes need to be checked if they are present at every single time step via ``hasattr()``
- when we want to reset the layer, we need to decide if we reset to 0 or to the initial state

Thus the question is: is there a real need for this or is it just for completeness? E.g., Lasagne does not learn these initial states per default.

fdlm:
We will have to deal with this problem anyways, as soon as we want to implement any streaming-capable algorithms for e.g. HMMs or CRFs, because they *always* have initial distributions/factors.

[...]

PR #233, opened 2016-12-02
[WIP] added initial states to RNN and LSTM layers

fdlm:
This is a first implementation of initial states for RNN and LSTM layers. Needs some testing. Addresses #230.

### DIFF
--- a/madmom/ml/nn/layers.py
+++ b/madmom/ml/nn/layers.py
@@ -98,12 +98,16 @@ class RecurrentLayer(FeedForwardLayer):
         Recurrent weights.
     activation_fn : numpy ufunc
         Activation function.
+    hid_init : numpy array, shape (), optional
+        Initial state of hidden units.
 
     """
 
-    def __init__(self, weights, bias, recurrent_weights, activation_fn):
+    def __init__(self, weights, bias, recurrent_weights, activation_fn,
+                 hid_init=None):
         super(RecurrentLayer, self).__init__(weights, bias, activation_fn)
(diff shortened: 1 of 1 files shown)


madmom/ml/nn/layers.py
```diff
@@ -283,16 +290,22 @@ class LSTMLayer(Layer):
...
     def __init__(self, input_gate, forget_gate, cell, output_gate,
-                 activation_fn=tanh):
+                 activation_fn=tanh, out_init=None, state_init=None):
         self.input_gate = input_gate
         self.forget_gate = forget_gate
         self.cell = cell
         self.output_gate = output_gate
         self.activation_fn = activation_fn
+        self.out_init = out_init
+        self.state_init = state_init
```
superbock: As above, I'm not really happy with the names here.
Also, for streaming mode, the output of the previous step must be accessible anyways, e.g. as ``self.out``.

For the ``state_init`` I am thinking of moving this as ``init`` to the cell, and store the cell's state there -- this is basically what it is. This needs some change in the ``activate()`` method, but is the cleaner way to solve this (IMO).

fdlm:
Since some of your proposed changes require more refactoring, I guess it would be really better to do this together with #185. It will be less work.

superbock:
Yes, I will work on this next week, since we need NNs in #185 anyways.

superbock:
Please see #235 for my first draft. If that PR addresses all your points we can close this PR.

superbock:
Closing, since #243 is merged

PR #240, opened 2016-12-19
[WIP] Online gmm bar tracking

superbock:
creating this PR, since I had a look at the code and want to comment on some stuff

### DIFF
--- a/madmom/audio/spectrogram.py
+++ b/madmom/audio/spectrogram.py
@@ -1115,7 +1115,7 @@ class SpectrogramDifferenceProcessor(BufferProcessor):
 
     def __init__(self, diff_ratio=DIFF_RATIO, diff_frames=DIFF_FRAMES,
                  diff_max_bins=DIFF_MAX_BINS, positive_diffs=POSITIVE_DIFFS,
-                 stack_diffs=None, sum_diffs=False, **kwargs):
+                 stack_diffs=None, **kwargs):
         # pylint: disable=unused-argument
         super(SpectrogramDifferenceProcessor, self).__init__(diff_frames)
         self.diff_ratio = diff_ratio
@@ -1123,7 +1123,6 @@ class SpectrogramDifferenceProcessor(BufferProcessor):
         self.diff_max_bins = diff_max_bins
         self.positive_diffs = positive_diffs
         self.stack_diffs = stack_diffs
-        self.sum_diffs = sum_diffs
 
     def process(self, data, **kwargs):
         """
@@ -1187,8 +1186,6 @@ class SpectrogramDifferenceProcessor(BufferProcessor):
             # TODO: check if this is still true, even with the buffer
             diff = self.stack_diffs((diff.spectrogram[self.diff_frames:],
                                      diff))
-        if self.sum_diffs:
-            diff = diff.sum()
         return diff
 
     @staticmethod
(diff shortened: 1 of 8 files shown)


madmom/audio/spectrogram.py
```diff
@@ -1115,14 +1115,15 @@ class SpectrogramDifferenceProcessor(BufferProcessor):
...
-                 stack_diffs=None, **kwargs):
+                 stack_diffs=None, sum_diffs=False, **kwargs):
         # pylint: disable=unused-argument
         super(SpectrogramDifferenceProcessor, self).__init__(diff_frames)
         self.diff_ratio = diff_ratio
         self.diff_frames = diff_frames
         self.diff_max_bins = diff_max_bins
         self.positive_diffs = positive_diffs
         self.stack_diffs = stack_diffs
+        self.sum_diffs = sum_diffs
```
superbock: This is not the right place to do summing of the differences -- nor is there a need to have an option for this.

You can simply add ``np.sum`` as a processor to the processing chain wherever you wanted to use this option.

madmom/features/beats_hmm.py
```diff
@@ -408,31 +408,41 @@ class MultiPatternTransitionModel(TransitionModel):
     transition_lambda : float, optional
         Lambda for the exponential tempo change distribution (higher values
         prefer a constant tempo from one pattern to the next one).
+    pattern_change_prob : float, optional
+        Probability of a pattern change. With pattern_change_prob - 1 we
+        maintain the old pattern.
```
superbock: The description does not match the parameter default value of 0.

madmom/features/beats_hmm.py
```diff
@@ -408,31 +408,41 @@ class MultiPatternTransitionModel(TransitionModel):
...
         # save attributes
         self.transition_models = transition_models
         self.transition_prob = transition_prob
         self.transition_lambda = transition_lambda
+        num_patterns = len(self.transition_models)
+        first_pattern_states = np.zeros(num_patterns, dtype=int)
+        last_pattern_states = np.zeros(num_patterns, dtype=int)
         # stack the pattern transitions
         for i, tm in enumerate(self.transition_models):
+
```
superbock: Please try to avoid inserting random empty lines.

madmom/features/beats_hmm.py
```diff
@@ -462,6 +495,10 @@ class RNNBeatTrackingObservationModel(ObservationModel):
     observation_lambda : int
         Split one beat period into `observation_lambda` parts, the first
         representing beat states and the remaining non-beat states.
+    obslik_floor : float
+        Observation likelihood floor. Values below this threshold and
+        above 1-threshold are clipped.
+
```
superbock: I think ``clip`` is a better name, since it does exactly this. We could also think of allowing a tuple with individual lower and upper clipping values.

Again, please do not insert additional empty lines.

madmom/features/beats_hmm.py
```diff
@@ -499,6 +538,12 @@ def log_densities(self, observations):
             Log densities of the observations.
 
         """
+        if self.obslik_floor is not None:
+            # limit the range of the observation likelihood to avoid
+            # problems if the probability gets 0 (impossible state) and 1 (
+            # obligatory state).
+            observations = np.maximum(observations, self.obslik_floor)
+            observations = np.minimum(observations, 1. - self.obslik_floor)
```
superbock: There's a shorthand for these two lines: ``np.clip()``.

madmom/ml/hmm.pyx
```diff
@@ -537,12 +584,8 @@ class HiddenMarkovModel(object):
...
 
         # init forward variables
-        if init is None:
-            for state in range(num_states):
-                fwd[0, state] = self.initial_distribution[state]
-        else:
-            for state in range(num_states):
-                fwd[0, state] = init[state]
+        for state in range(self.initial_distribution.shape[0]):
+            fwd[0, state] = self.initial_distribution[state]
```
superbock: Please revert these changes, since this change is required by other code to work.

NB: it is likely that the ``init`` parameter will be removed later on, since it is a very dirty way of keeping the state of the HMM, but right now it is required.

PR #243, opened 2017-01-18
[WIP] stateful RNNs

superbock:
This is the latest attempt so solve #230 and #234. It also address partly #185, at least the parts relevant to NNs.

Supersedes #233 and #235.

### DIFF
--- a/madmom/ml/nn/__init__.py
+++ b/madmom/ml/nn/__init__.py
@@ -57,11 +57,11 @@ class NeuralNetwork(Processor):
 
     >>> from madmom.ml.nn.layers import FeedForwardLayer
     >>> from madmom.ml.nn.activations import tanh, sigmoid
-    >>> l1_weights = [[0.5, -1., -0.3 , -0.2]]
-    >>> l1_bias = [0.05, 0., 0.8, -0.5]
+    >>> l1_weights = np.array([[0.5, -1., -0.3 , -0.2]])
+    >>> l1_bias = np.array([0.05, 0., 0.8, -0.5])
     >>> l1 = FeedForwardLayer(l1_weights, l1_bias, activation_fn=tanh)
-    >>> l2_weights = [-1, 0.9, -0.2 , 0.4]
-    >>> l2_bias = [0.5]
+    >>> l2_weights = np.array([-1, 0.9, -0.2 , 0.4])
+    >>> l2_bias = np.array([0.5])
     >>> l2 = FeedForwardLayer(l2_weights, l2_bias, activation_fn=sigmoid)
     >>> nn = NeuralNetwork([l1, l2])
     >>> nn  # doctest: +ELLIPSIS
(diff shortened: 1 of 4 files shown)


madmom/ml/nn/__init__.py
```diff
@@ -71,8 +71,17 @@ class NeuralNetwork(Processor):
...
 
-    def __init__(self, layers):
+    def __init__(self, layers, online=False):
         self.layers = layers
+        self.online = online
+
+    def __setstate__(self, state):
+        # restore instance attributes
+        self.__dict__.update(state)
+        # TODO: old models do not have the online attribute, thus create it
```
fdlm: Isn't this an explanation, not a "TODO"?

madmom/ml/nn/__init__.py
```diff
@@ -71,8 +71,17 @@ class NeuralNetwork(Processor):
...
+        #       remove this initialisation code after updating the models
+        if not hasattr(self, 'online'):
+            self.online = None
```
fdlm: Why is in initialised with `None`, and not `False`?

madmom/ml/nn/layers.py
```diff
@@ -23,9 +23,16 @@ class Layer(object):
 
     """
 
-    def __call__(self, *args):
+    def __call__(self, *args, **kwargs):
         # this magic method makes a Layer callable
-        return self.activate(*args)
+        return self.activate(*args, **kwargs)
+
+    def __getstate__(self):
```
fdlm: IMHO, this shouldn't be in the `Layer` class. Every subclass of `Layer` should be responsible to ensure that their state is not pickled. This does not lead to much code duplication - right now, only `RecurrentLayer` and `LSTMLayer` are stateful, and I think for `LSTMLayer` you'll have to overwrite `__getstate__` anyways, because you don't want to pickle the previous state and the cell state.

madmom/ml/nn/layers.py
```diff
@@ -44,16 +51,24 @@ def activate(self, data):
         """
         raise NotImplementedError('must be implemented by subclass.')
 
+    @staticmethod
```
fdlm: Not sure this should be a static method - sure, it does not change the object state in the `Layer` class, but its functionality is **intended** to change the object state.

madmom/ml/nn/layers.py
```diff
@@ -90,51 +108,105 @@ class RecurrentLayer(FeedForwardLayer):
...
+        Reset the layer to its initial state.
+
+        Parameters
+        ----------
+        init : numpy array, shape (num_hiddens,), optional
+            Reset the hidden units to this initial state.
+
+        """
+        # reset previous time step to initial value
+        self._prev = self.init if init is None else init
```
fdlm: `self._prev = init or self.init` ?

madmom/ml/nn/layers.py
```diff
@@ -283,68 +364,146 @@ class LSTMLayer(Layer):
...
+        # TODO: old models do not have the init attributes, thus create them
+        #       remove this initialisation code after updating the models
+        if not hasattr(self, 'init'):
+            self.init = np.zeros(self.cell.bias.size, dtype=NN_DTYPE)
+        if not hasattr(self, 'cell_init'):
+            self.cell_init = np.zeros(self.cell.bias.size, dtype=NN_DTYPE)
+        # add non-pickled attributes needed for stateful processing
+        self._prev = self.init
+        self._state = self.cell_init
+
```
fdlm: Shouldn't `LSTMLayer` also have a `__getstate__` for filtering out `self._prev` **and** `self._state` for pickling?

madmom/ml/nn/layers.py
```diff
@@ -413,7 +572,7 @@ def activate(self, data, reset_gate, prev):
         return self.activation_fn(out)
```
fdlm: This functionality looks very similar to what  the `Gate` class already provides. Do we really need a separate class for the GRU cell?

madmom/ml/nn/layers.py
superbock: Yes, this was a normal method before, changed it because some checker criticised it. Will change it back.

madmom/ml/nn/layers.py
superbock: 👍

madmom/ml/nn/layers.py
superbock: Unfortunately yes, since it is only similar...

madmom/ml/nn/__init__.py
superbock: see below...

madmom/ml/nn/layers.py
```diff
@@ -445,14 +611,27 @@ class GRULayer(Layer):
...
-        if hid_init is None:
-            hid_init = np.zeros(cell.bias.size, dtype=NN_DTYPE)
-        self.hid_init = hid_init
+        # keep the state of the layer
+        if init is None:
+            init = np.zeros(self.cell.bias.size, dtype=NN_DTYPE)
+        self.init = init
+        # keep the state of the layer
+        self._prev = self.init
+
```
fdlm: Missing a `__getstate__` here, because it's no longer in `Layer`. (I still think it's good to have it here instead of there)

Issue #239, opened 2016-12-08
forward method of HMMs is not stateful


superbock:
If ``HMM.forward`` is called multiple times with different input data, it forgets its state between calls, because it is initialised every time.

Either (i) we keep the state and add a flag, whether it should be reseted or not (e.g. via ``reset``), or (ii) we add an option to pass the previous state instead of initialising it every time.

I vote for (i), since it is definitely more object-oriented, but (ii) might be easier to implement or requires less changes.

PR #244, opened 2017-01-20
Stateful processing

superbock:
This PR adds proper stateful processing.

The `process()` method of stateful processors accepts an additional/new `reset` argument, which defaults to `True` and thus does not change existing behaviour. This is implemented for `HMMs` and `RNNs`.

To be able to process sequences either as a whole or on a frame-by-frame basis, `process()` expects always the same data dimensionality. `NeuralNetwork` previously corrected the dimensions if a 1-dimensional input feature is expected (e.g. `SuperFluxNN`). This conversion is moved to the executable program instead.

If data is processed framewise, `reset=False` must be passed to the processors, otherwise stateful processors get reseted constantly. To be able to pass additional keyword arguments, `madmom.processors` was adapted accordingly.

### DIFF
--- a/bin/SuperFluxNN
+++ b/bin/SuperFluxNN
@@ -8,6 +8,7 @@ SuperFlux with neural network based peak picking onset detection algorithm.
 from __future__ import absolute_import, division, print_function
 
 import argparse
+import numpy as np
 
 from madmom.processors import IOProcessor, io_arguments
 from madmom.audio.signal import SignalProcessor, FramedSignalProcessor
@@ -103,7 +104,9 @@ def main():
         pp = PeakPickingProcessor(**vars(args))
         from madmom.utils import write_events as writer
         # sequentially process them
-        out_processor = [rnn, pp, writer]
+        # Note: we need np.atleast_2d and np.transpose before the RNN, since
+        #       it expects the data in 2D (1D means framewise processing)
+        out_processor = [np.atleast_2d, np.transpose, rnn, pp, writer]
 
     # create an IOProcessor
     processor = IOProcessor(in_processor, out_processor)
--- a/madmom/ml/hmm.pyx
+++ b/madmom/ml/hmm.pyx
@@ -394,6 +394,34 @@ class HiddenMarkovModel(object):
             raise ValueError('Initial distribution is not a probability '
                              'distribution.')
         self.initial_distribution = initial_distribution
+        # attributes needed for stateful processing (i.e. forward_step())
+        self._prev = self.initial_distribution.copy()
+
+    def __getstate__(self):
+        # copy everything to a pickleable object
+        state = self.__dict__.copy()
+        # do not pickle attributes needed for stateful processing
+        state.pop('_prev', None)
+        return state
+
+    def __setstate__(self, state):
+        # restore pickled instance attributes
+        self.__dict__.update(state)
+        # add non-pickled attributes needed for stateful processing
+        self._prev = self.initial_distribution.copy()
+
+    def reset(self, initial_distribution=None):
+        """
+        Reset the HMM to its initial state.
+
+        Parameters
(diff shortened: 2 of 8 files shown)


superbock:
This fixes #239.

bin/SuperFluxNN
```diff
@@ -103,7 +104,9 @@ def main():
         pp = PeakPickingProcessor(**vars(args))
         from madmom.utils import write_events as writer
         # sequentially process them
-        out_processor = [rnn, pp, writer]
+        # Note: we need np.atleast_2d and np.transpose before the RNN, since
+        #       it expects the data in 2D (1D means framewise processing)
+        out_processor = [np.atleast_2d, np.transpose, rnn, pp, writer]
```
superbock: Yes and no. I thought about this as well, but opted for the way of least invasion to existing code.
As you state, returning a 2d-array is a bit unintuitive.

There's a TODO item in #185 which says: "rewrite all features as processors (mostly `madmom.features.onsets.*`)". I think it is better to address this separately and not in this PR.

madmom/ml/hmm.pyx
```diff
@@ -518,52 +550,61 @@ class HiddenMarkovModel(object):
...
+        # reset HMM
+        if reset:
+            self.reset()
 
         # forward variables
-        cdef double[:, ::1] fwd = np.zeros((num_observations + 1, num_states),
-                                           dtype=np.float)
+        cdef double[::1] fwd_prev = self._prev
+        cdef double[:, ::1] fwd_cur = \
+            np.zeros((num_observations, num_states), dtype=np.float)
```
superbock: 1. ok, this makes sense.
2. `fwd_cur` holds only `num_observations`, not `num_observations + 1` variables any more. The previous variables are stored in `self._prev`. This change was necessary (at least I think it is) because during initialisation we know how many states we have and can thus init `self._prev`, but don't know the lengths of the observations yet. If we want to be able to call `forward()` with observations of variable length (i.e. both a complete sequence or frame-by-frame), we have to init `fwd_cur` (or `fwd`, see 1.) inside `forward()` but need to keep the previous variables somewhere, hence `self._prev`.

madmom/ml/hmm.pyx
```diff
@@ -518,52 +550,61 @@ class HiddenMarkovModel(object):
...
-            # indices for current and previous time step
-            cur = frame + 1
-            prev = frame
             # keep track of the normalisation sum
             prob_sum = 0
             # iterate over all states
             for state in range(num_states):
+                # we need to reset the current state in case this method
+                # gets called multiple times (i.e. framewise processing)
+                fwd_cur[frame, state] = 0
```
superbock: Yes, this is a leftover from a previous code version. Back then I also initialised `fwd_cur` of length 1 during initialisation and only re-allocated it if called with observations with length > 1.

madmom/ml/hmm.pyx
```diff
@@ -394,6 +394,38 @@ class HiddenMarkovModel(object):
             raise ValueError('Initial distribution is not a probability '
                              'distribution.')
         self.initial_distribution = initial_distribution
+        # attributes needed for stateful processing (i.e. forward_step())
+        self._prev = self.initial_distribution.copy()
+        self._cur = np.zeros_like(self._prev)
```
superbock: Yes, calling `reset()` does the same.

IMHO, we should get rid of `forward_generator` altogether in the long run, since it basically does the same as `forward()` with `reset=False` called with individual observations. What the current implementation of `forward()` doesn't support is handling of `block_size` other than 1. Thus I did not remove it for now.

superbock: Ok, I opted against calling `reset()`, because I think it is more explicit to initialise variables in `__init__()` and not in any other method.

PR #260, opened 2017-03-01
Grammar

carlthome:


superbock:
Thanks. I will not merge it right now, otherwise I have to rebase a lot of other branches which will be merged soon. Also, the same mistake is done all over `/bin` since it is mostly copy paste, will correct them as well then.

PR #272, opened 2017-04-10
Handle absent “filterbank” attribute of Spectrogram in chroma calculations

jpauwels:
Correct me if I'm wrong, but a regular Spectrogram object, e.g. as instantiated two lines above, doesn't have a filterbank attribute, leading to failure on these two lines.


superbock:
No, you're absolutely right. This is definitely a leftover from refactoring the filterbank handling to have less overhead (especially for online processing) in 65177fd.

But it should be enough to simply check for the `filterbank` attribute since the base `Spectrogram` class does not have it anymore (as it did previously for historical reasons), or to check if a `FilteredSpectrogram` was passed.

However, I am not sure if we shouldn't just drop the check completely and use whatever spectrogram was passed.

superbock:
Superseded by #281, closing. Thanks for reporting!

PR #282, opened 2017-05-10
WIP: Mfccs

scriptator:
Unit tests still missing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,7 @@
 
 # PyCharm files
 /.idea/
+*.iml
 
 # MacOS files
 .DS_Store
--- a/madmom/audio/cepstrogram.py
+++ b/madmom/audio/cepstrogram.py
@@ -121,6 +121,7 @@ MFCC_FMAX = 15000.
 MFCC_NORM_FILTERS = True
 MFCC_MUL = 1.
 MFCC_ADD = np.spacing(1)
+MFCC_DCT_NORM = "ortho"
 
 
 class MFCC(Cepstrogram):
@@ -150,6 +151,8 @@ class MFCC(Cepstrogram):
         logarithm.
     add : float, optional
         Add this value before taking the logarithm of the magnitudes.
+    dct_norm : {None, 'ortho'}, optional
+        Normalization mode (see scipy.fftpack.dct). Default is 'ortho'.
     kwargs : dict
         If no :class:`.audio.spectrogram.Spectrogram` instance was given, one
         is instantiated and these keyword arguments are passed.
@@ -181,14 +184,14 @@ class MFCC(Cepstrogram):
     def __init__(self, spectrogram, filterbank=MelFilterbank,
                  num_bands=MFCC_BANDS, fmin=MFCC_FMIN, fmax=MFCC_FMAX,
                  norm_filters=MFCC_NORM_FILTERS, mul=MFCC_MUL, add=MFCC_ADD,
-                 **kwargs):
+                 dct_norm=MFCC_DCT_NORM, **kwargs):
         # this method is for documentation purposes only
         pass
 
     def __new__(cls, spectrogram, filterbank=MelFilterbank,
                 num_bands=MFCC_BANDS, fmin=MFCC_FMIN, fmax=MFCC_FMAX,
                 norm_filters=MFCC_NORM_FILTERS, mul=MFCC_MUL, add=MFCC_ADD,
-                **kwargs):
+                dct_norm=MFCC_DCT_NORM, **kwargs):
         # for signature documentation see __init__()
         from .filters import Filterbank
         # instantiate a Spectrogram if needed
@@ -210,8 +213,8 @@ class MFCC(Cepstrogram):
         data = np.dot(spectrogram, filterbank)
         # logarithmically scale the magnitudes
         np.log10(mul * data + add, out=data)
-        # apply DCT
-        data = dct(data)
+        # apply type 2 DCT
+        data = dct(data, norm=dct_norm)
         # cast as MFCC
         obj = np.asarray(data).view(cls)
         # save additional attributes
--- a/madmom/audio/cepstrogram.py
+++ b/madmom/audio/cepstrogram.py
@@ -19,6 +19,7 @@ from scipy.fftpack import dct
 from .filters import MelFilterbank
 from .spectrogram import Spectrogram
 from ..processors import Processor
+from ..utils import lazyprop
 
 
 class Cepstrogram(np.ndarray):
@@ -37,7 +38,6 @@ class Cepstrogram(np.ndarray):
         one is instantiated with these additional keyword arguments.
 
     """
-
     # pylint: disable=super-on-old-class
     # pylint: disable=super-init-not-called
     # pylint: disable=attribute-defined-outside-init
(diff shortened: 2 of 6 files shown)


.gitignore
```diff
@@ -27,6 +27,7 @@
 
 # PyCharm files
 /.idea/
+*.iml
```
superbock: Not sure if `.iml` files are that common. I'd rather have them not included in `.gitignore`.

madmom/audio/cepstrogram.py
```diff
@@ -34,6 +37,7 @@ class applies some transformation (usually a DCT) on a spectrogram.
         one is instantiated with these additional keyword arguments.
 
     """
+
```
superbock: Please try avoiding adding/changing whitespace/empty lines.

madmom/audio/cepstrogram.py
```diff
@@ -121,6 +125,22 @@ def process(self, data, **kwargs):
...
+def lazyprop(fn):
+    attr_name = '_lazy_' + fn.__name__
+
+    @property
+    def _lazyprop(self):
+        if not hasattr(self, attr_name):
+            setattr(self, attr_name, fn(self))
+        return getattr(self, attr_name)
+
+    return _lazyprop
```
superbock: This decorator can go into utils since it is rather useful. Please document it, though.

madmom/audio/cepstrogram.py
```diff
@@ -150,6 +170,8 @@ class MFCC(Cepstrogram):
         logarithm.
     add : float, optional
         Add this value before taking the logarithm of the magnitudes.
+    dct_norm : {None, 'ortho'}, optional
+        Normalization mode (see scipy.fftpack.dct). Default is 'ortho'.
```
superbock: if 'ortho' is the default, it should go first

madmom/audio/cepstrogram.py
```diff
@@ -174,21 +196,22 @@ class MFCC(Cepstrogram):
     5) The MFCCs are the amplitudes of the resulting spectrum
 
     """
+
```
superbock: see above

madmom/audio/cepstrogram.py
```diff
@@ -197,7 +220,7 @@ def __new__(cls, spectrogram, filterbank=MelFilterbank,
             spectrogram = Spectrogram(spectrogram, **kwargs)
 
         # instantiate a Filterbank if needed
-        if issubclass(filterbank, Filterbank):
+        if inspect.isclass(filterbank) and issubclass(filterbank, Filterbank):
```
superbock: Can you please give an example of when this check is needed? If it is a common case, we should consider the same logic at other places as well (e.g. `FilteredSpectrogram`).

madmom/audio/cepstrogram.py
```diff
@@ -222,6 +245,63 @@ def __new__(cls, spectrogram, filterbank=MelFilterbank,
...
 
+    @staticmethod
+    def calc_deltas(data, delta_filter):
+        """
+        Applies the given filter to the data after automatically padding by
+        replicating the first and last frame. The length of the padding is
+        calculated via ceil(len(delta_filter))
+        :param data: a numpy array with the data
+        :param delta_filter: a numpy array containing the filter
+        :return: a numpy array of the same shape as data, containing the deltas
```
superbock: Please use [numpydoc](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt) style docstrings. Also, try following [PEP257](https://www.python.org/dev/peps/pep-0257/) (e.g. "Apply the...").

madmom/audio/cepstrogram.py
```diff
@@ -222,6 +245,63 @@ def __new__(cls, spectrogram, filterbank=MelFilterbank,
...
+
+    @lazyprop
+    def deltas(self, delta_filter=MFCC_DELTA_FILTER):
+        return MFCC.calc_deltas(self, delta_filter)
+
+    @lazyprop
+    def deltadeltas(self, deltadelta_filter=MFCC_DELTADELTA_FILTER):
+        return MFCC.calc_deltas(self.deltas, deltadelta_filter)
+
+    def calc_ml_deltas(self, delta_filter=MFCC_DELTA_FILTER,
```
superbock: I kind of dislike the abbreviation "ml" for Matlab, since it is used for machine learning elsewhere in the package.

madmom/audio/signal.py
```diff
@@ -496,13 +496,10 @@ def load_wave_file(filename, sample_rate=None, num_channels=None, start=None,
...
     if sample_rate is not None and sample_rate != file_sample_rate:
-        raise ValueError('Requested sample rate of %f Hz, but got %f Hz and '
-                         're-sampling is not implemented.' %
-                         (sample_rate, file_sample_rate))
+        signal = resample(signal, sample_rate=sample_rate)
     # same for the data type
     if dtype is not None and signal.dtype != dtype:
-        raise ValueError('Requested dtype %s, but got %s and re-scaling is '
-                         'not implemented.' % (dtype, signal.dtype))
+        signal = rescale(signal, dtype=dtype)
```
superbock: You forgot to change the tests accordingly, tests check for these errors.

scriptator:
I will incorporate the requested changes, no problem. Thanks for the detailed review!

.gitignore
scriptator: I am working with IntelliJ, which needs and always produces this iml file for storing the module information. It may be less common than PyCharm but I don't see the benefit of installing PyCharm as well, given that the main difference between the two programs is the list of preinstalled plugins. I simply installed the python plugin to my IntelliJ and have a a PyCharm clone.

If that does not convince you we can leave it out though, I just wanted to explain how it comes that I added the line.

madmom/audio/cepstrogram.py
scriptator: My code formatter is doing this, I will try to change the behaviour. Else it will happen again for sure.

madmom/audio/cepstrogram.py
scriptator: Just pass in a filterbank instance. Then, the first argument of issubclass won't be a class which provokes the following error:

`TypeError: issubclass() arg 1 must be a class`

For example, in my evaluation script I am doing it like that, which causes an error without my fix:
> filterbank = Filterbank(filterdata, pyfft.bin_frequencies)
> mfcc = MFCC(pyfft, filterbank=filterbank, num_bands=no_filters, fmin=0, fmax=fs / 2)

madmom/audio/signal.py
scriptator: In your opinion, does it make sense to write tests asserting that the resampling in the context of `load_wave_file()` works correctly? The resample function itself is already tested quite well

superbock: No, but right now the tests test for this ValueError to be raised. This needs to changed accordingly. Please note that [some doctests fail as well](https://travis-ci.org/CPJKU/madmom/jobs/231804552#L658), but I'm not sure if this is related or not.

scriptator: I figured it out now. My assumption was that I wouldn't ruin anything by replacing these ValueErrors by the implementation because nobody could have used that function for loading something which requires resampling or rescaling anyways. However, I was wrong... the `load_audio_file` function in signal.py first tries to load using `load_wave_file` and falls back to ffmpeg if an Error is thrown. Now, as the latter function doesn't return errors any more it loads, resamples and rescales correctly what ffmpeg did before, only that the results differ in some late decimal places. I guess that explains the failing doctests as well.

What would you do? Maybe opening a separate issue would be the right solution, as my changes turned out not to be as simple as I assumed them to be. I don't need them for my branch, so I could revert the commit.

superbock: Yes, creating a separate issue and skipping that commit (when rebasing the whole branch at the end) is the cleanest solution.

scriptator:
I added some unit tests for MFCC, please tell me if you miss something important? For some problem with np.vectorize, python 3.4 is failing. Is the numpy version different there?

madmom/audio/cepstrogram.py
superbock: Just figured out that this is exactly what's done in `audio.spectrogram` ;)

scriptator: Haha, well then I guess I can leave it like that?

Issue #283, opened 2017-05-16
TransitionModel returns wrong number of states if state is unreachable


fdlm:
In this example, the last state is not reachable:
```python
>>> A = np.array([[.5, .5, 0.], [.5, .5, 0.], [.5, .5, 0.]])
>>> frm, to = A.nonzero()
>>> tm = TransitionModel.from_dense(to, frm, A[frm, to])
>>> print tm.num_states
2
```
Expected output would be '3'.

This is because `num_states` in `TransitionModel` relies on the length of `self.pointers`. A possible solution might be to use `states.max() + 1`.

fdlm:
fixed by #287

PR #284, opened 2017-05-16
Fixed num_states property in TransitionModel

fdlm:
It should now report the correct number of states, even if some of them are unreachable. Addresses #283.

However, accessing `num_states` now takes longer than before. Should we cache this value?

### DIFF
--- a/madmom/ml/hmm.pyx
+++ b/madmom/ml/hmm.pyx
@@ -82,11 +82,12 @@ class TransitionModel(object):
         self.states = states
         self.pointers = pointers
         self.probabilities = probabilities
+        self._num_states = int(states.max() + 1)
 
     @property
     def num_states(self):
         """Number of states."""
-        return len(self.pointers) - 1
+        return self._num_states
 
     @property
     def num_transitions(self):
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,10 @@ Release Notes
 Version 0.16.dev0
 -----------------
 
+Bug fixes:
+
+* Fix `TransitionModel` not returning the correct number of states (#284)
+
 Other changes:
 
 * Viterbi decoding of `HMM` raises a warning if no valid path is found (#279)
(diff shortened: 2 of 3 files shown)


superbock:
Some tests for Python 3 fail ("AssertionError: 30 is not an instance of <class 'int'>"), either we change the tests or cast it to int.

superbock:
One more question regarding the longer access times. How much longer does it take? Since `num_states` is accessed only once in `viterbi()`/`forward()` it seems not to be crucial though. Even in online mode where `forward()` is called every frame it seems not to slow down things much.

fdlm:
I did not experience a slowdown, but in theory, for each call, it has to run over all transitions to find the max value. Since storing a single integer does not cost much memory, I'd go with storing it. It just makes more sense to me.

superbock:
Theoretically, people can overwrite it then, thus I went with properties instead of storing an attribute. However, people doing this should be aware of the consequences anyways. I just tested the current property with the `.max()` with `DBNBeatTracker` in online mode and has exactly the same speed as before. So need to bother, but do as you like.

fdlm:
To clarify, do I add the PR number or the Issue number to CHANGES.rst?

superbock:
I usually add the PR number.

madmom/ml/hmm.pyx
```diff
@@ -427,6 +428,113 @@ class HiddenMarkovModel(object):
...
+            path[frame] = state
+            # fetch the next previous one
+            state = bt_pointers[frame, state]
+
+        # return the tracked path and its probability
+        return path, log_probability, np.asarray(bt_pointers), np.asarray(viterbi_vars)
+
+    @cython.cdivision(True)
+    @cython.boundscheck(False)
+    @cython.wraparound(False)
```
superbock: Is adding this `verbose_viterbi()` a mistake or on purpose? If on purpose, I'd propose to add a verbose parameter to `viterbi()` instead.

fdlm: No, I messed up the commit. I'm going to clean this up.

fdlm:
I'll close this PR, because it fixes only a symptom of another problem (see #285).

Issue #288, opened 2017-05-17
AppVeyor ffmpeg download: Not found


SebastianPoell:
Hi,

AppVeyor tries to download ffmpeg-2.7 and fails because the version 2.7 does not exist (anymore):
https://ffmpeg.zeranoe.com/builds/win64/shared/

![screen shot 2017-05-17 at 11 22 28](https://cloud.githubusercontent.com/assets/1398398/26147225/2a25d280-3af3-11e7-8cdd-6745affcb310.png)

I think that can easily be fixed by using another ffmpeg version in appveyor.yml. E.g:
```
- ps: Start-FileDownload ('http://ffmpeg.zeranoe.com/builds/win' + $env:PYTHON_ARCH + '/shared/ffmpeg-latest-win' + $env:PYTHON_ARCH + '-shared.7z' ) ffmpeg-shared.7z
```

Regards, Sebi

superbock:
PR welcome :)

Issue #178, opened 2016-07-20
fix stuck Appveyor builds


superbock:
All of a sudden, Appveyor builds got stuck. Needs investigation. See https://github.com/appveyor/ci/issues/923 for more details.

PR #289, opened 2017-05-17
Fix for #288

SebastianPoell:
Could not test it but should work.


SebastianPoell:
Ok the ffmpeg issue is fixed by this PR.
The Endless Loop Issue #178 remains :(

PR #286, opened 2017-05-16
Online TempoDetector: (Comb, ACF, DBN)

SebastianPoell:
Hey,

Here my first Pull Request. This should work now (inspired by the online version of the DBNBeatTracker):

python3 TempoDetector -v online
python3 TempoDetector online /wathever.wav
python3 TempoDetector single --online /whatever.wav

Update: Also added support for online acf and dbn:
python3 TempoDetector --method acf -v online
python3 TempoDetector --method dbn -v online

Let me know if there are things to change :)

Regards, Sebi

PR #294, opened 2017-06-07
[WIP] Online Tempo Estimators

SebastianPoell:
This is a refactoring of PR #286 based on the refactor_tempo branch.

### DIFF
--- a/madmom/features/tempo.py
+++ b/madmom/features/tempo.py
@@ -645,11 +645,12 @@ class CombFilterTempoEstimationProcessor(BaseTempoEstimationProcessor):
     alpha : float, optional
         Scaling factor for the comb filter.
     buffer_size : int, optional
-        For online mode a buffer [seconds] for the summed up bins is used.
+        Use a buffer with `buffer_size` seconds for online mode
+        to sum up the bins.
     fps : float, optional
         Frames per second.
     online : bool, optional
-        Extends the combfilter matrix frame by frame.
+        Extend the combfilter matrix frame by frame.
 
     Examples
     --------
@@ -816,11 +817,11 @@ class ACFTempoEstimationProcessor(BaseTempoEstimationProcessor):
     alpha : float, optional
         Scaling factor for the comb filter.
     buffer_size : int, optional
-        For online mode a buffer [seconds] for the activations is used.
+        Buffer activations over `buffer_size` seconds for online mode.
     fps : float, optional
         Frames per second.
     online : bool, optional
-        Uses only the buffered activations to perform the auto correlation.
+        Use only the buffered activations to perform the auto correlation.
 
     Examples
     --------
@@ -929,7 +930,7 @@ class DBNTempoEstimationProcessor(BaseTempoEstimationProcessor):
     fps : float, optional
         Frames per second.
     online : bool, optional
-        Uses the forward algorithm to retrieve the tempo.
+        Use the forward algorithm to retrieve the tempo.
 
     Examples
     --------
--- a/madmom/features/tempo.py
+++ b/madmom/features/tempo.py
@@ -12,6 +12,7 @@ from __future__ import absolute_import, division, print_function
 import numpy as np
 import sys
 
+from .beats import DBNBeatTrackingProcessor
 from madmom.processors import Processor, BufferProcessor
 from madmom.audio.signal import smooth as smooth_signal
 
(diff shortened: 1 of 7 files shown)


SebastianPoell:
Regarding the verbose online mode: 
After searching the web it seems like refreshing multiple lines in the console requires external libraries (maybe even different libraries depending on the operating system). So for now I just went with an one-line representation: The most likely 3 tempi and their strength.

screen shot 2017-06-09 at 00 53 31

SebastianPoell:
@superbock: I think it's much cleaner now. Would be awesome if you could look through all of this if you have time :)

madmom/features/tempo.py
```diff
@@ -10,8 +10,10 @@
 from __future__ import absolute_import, division, print_function
 
 import numpy as np
+import sys
 
-from madmom.processors import Processor
+from .beats import DBNBeatTrackingProcessor
```
superbock: Please import only inside the needed class.

madmom/features/tempo.py
```diff
@@ -764,14 +960,23 @@ class DBNTempoEstimationProcessor(BaseTempoEstimationProcessor):
...
         # pylint: disable=unused-argument
         super(DBNTempoEstimationProcessor, self).__init__(
             min_bpm=min_bpm, max_bpm=max_bpm, act_smooth=act_smooth,
-            hist_smooth=hist_smooth, fps=fps, **kwargs)
+            hist_smooth=hist_smooth, fps=fps, online=online, **kwargs)
+        # save additional variables
+        if self.online:
+            self._dbn = DBNBeatTrackingProcessor(min_bpm=self.min_bpm,
+                                                 max_bpm=self.max_bpm,
+                                                 fps=self.fps)
```
superbock: I would create the `DBNBeatTrackingProcessor` as `self.dbn` not only in online, but also in offline mode and use change `interval_histogram()` accordingly. While doing this, it would be good to also pass `**kwargs` instead of setting `num_tempi=None`.

The `from .beats import DBNBeatTrackingProcessor` should go in here rather than at the module level.

SebastianPoell: Very nice suggestions, thx!
Implemented here: 0b8ea8a883b111988add80d0332fb624afbed1e8

madmom/features/tempo.py
```diff
@@ -720,6 +891,29 @@ def interval_histogram(self, activations):
...
+            Bins of the tempo histogram.
+        histogram_delays : numpy array
+            Corresponding delays [frames].
+        """
+        # shift buffer and put new activation at end of buffer
+        buffer = self._buffer(activation)
+        # use offline acf function on buffered activations
+        return interval_histogram_acf(buffer, self.min_interval,
+                                      self.max_interval)
+
```
SebastianPoell: Yeah, I have to test the ACF performance anyway. If it performs similar to the other methods I will try to change it into a "real" online computation.

madmom/features/tempo.py
```diff
@@ -575,15 +683,27 @@ class CombFilterTempoEstimationProcessor(BaseTempoEstimationProcessor):
     HIST_SMOOTH = 9
     ACT_SMOOTH = 0.14
     ALPHA = 0.79
+    BUFFER_SIZE = 1000
```
SebastianPoell: Absolutely right. Changed here: 7ead132d7e1d4770e9f338fd2b869e1b992eb273

madmom/features/tempo.py
```diff
@@ -310,6 +335,64 @@ def process(self, activations, **kwargs):
...
+
+        Returns
+        -------
+        tempi : numpy array
+            Array with the dominant tempi [bpm] (first column) and their
+            relative strengths (second column).
+
+        """
+        # make the activations a 1D array
+        activations = np.atleast_1d(activations)
```
SebastianPoell: Ah ok, I removed it now and adapted the tests (like you did for #299) here:
25a7522eda79907cdb07e5945214b89449f803b3

madmom/features/tempo.py
SebastianPoell: Also included in this commit: 0b8ea8a883b111988add80d0332fb624afbed1e8

madmom/features/tempo.py
```diff
@@ -543,8 +647,12 @@ class CombFilterTempoEstimationProcessor(BaseTempoEstimationProcessor):
         Smooth the tempo histogram over `hist_smooth` bins.
     alpha : float, optional
         Scaling factor for the comb filter.
+    buffer_size : int, optional
+        For online mode a buffer [bins] for the summed up bins is used.
     fps : float, optional
         Frames per second.
+    online : bool, optional
+        Extends the combfilter matrix frame by frame.
```
SebastianPoell: Hope thats ok now: e4babd14cd5873f24394dd1ede7e9ef7312b3ee2

madmom/features/tempo.py
```diff
@@ -605,6 +725,44 @@ def interval_histogram(self, activations):
...
+        act_max = self.combfilter_matrix[-1] == \
+            np.max(self.combfilter_matrix[-1], axis=-1)
+        # compute the max bins
+        bins = self.combfilter_matrix[-1] * act_max
+        # use a buffer to only keep bins of the last seconds
+        # shift buffer and put new bins at end of buffer
+        buffer = self._buffer(bins)
+        # build a histogram together with the intervals and return it
+        return np.sum(buffer, axis=0), np.array(self.taus)
+
```
SebastianPoell: Note: if I do a simple measurement of the performance with `time.time()` for the whole `online_interval_histogram()` method I get an average execution time of almost exactly 1ms (on my 2013 MacBook Air).

superbock:
Thanks, I will merge it now and work on it a bit further. Nice work!

PR #297, opened 2017-06-09
Fixed BeatTracker double detections

SebastianPoell:
The BeatTracker sometimes yield double detections which then result in errors when evaluating.
I just added an additional _np.unique_ before returning the detections.

If further investigation is needed I can send activations which will produce such double detections.

Regards, Sebi

### DIFF
--- a/madmom/features/beats.py
+++ b/madmom/features/beats.py
@@ -532,6 +532,8 @@ class BeatTrackingProcessor(Processor):
 
         # convert detected beats to a list of timestamps
         detections = np.array(detections) / float(self.fps)
+        # make detections unique
+        detections = np.unique(detections)
         # remove beats with negative times and return them
         return detections[np.searchsorted(detections, 0):]
         # only return beats with a bigger inter beat interval than that of the


superbock:
Yes, this would definitely help to understand the problem. Although the proposed change may fix the evaluation problem, it will definitely not fix the underlying problem.

SebastianPoell:
Sure, here are two activation files. One of them leeds to a double detection in the beginning and one of them in the end.

[detections.zip](https://github.com/CPJKU/madmom/files/1064261/detections.zip)

Regards, Sebi

superbock:
I had a look into the issue and discovered that this situation arose if the detected interval in the activation function changes. This can occur due to the iterative approach of detecting the beats searching for the next one at and around the position given the last detected beat and the dominant interval. If the interval for the next block changes, the same beat can be detected multiple times. I think #298 fixes it in a better way by requiring the net beat to be greater than the last detected one plus the minimum interval given by the max. tempo.

SebastianPoell:
Thank you for looking into this in more detail!
Yeah, I also suspected interval changes and your solution looks good to me 👍 

So, I think this can be closed?

PR #301, opened 2017-06-26
Made ACF capable of real online processing

SebastianPoell:
ACF and Comb seem to perform very similar for beat tracking. To save time for future evaluations I wanted to optimize it sooner rather than later ;)

Instead of executing the ACF on the buffered activations the method now processes the ACF in real-time. Changing the tests was not necessary since this produces the exact same results as before. 
I also did some basic performance tests and this is **way** faster.

Regards, Sebi

### DIFF
--- a/madmom/features/tempo.py
+++ b/madmom/features/tempo.py
@@ -288,6 +288,11 @@ class BaseTempoEstimationProcessor(OnlineProcessor):
         """Maximum beat interval [frames]."""
         return int(np.ceil(60. * self.fps / self.min_bpm))
 
+    @property
+    def taus(self):
+        """List of all lag times from min to max interval."""
+        return np.arange(self.min_interval, self.max_interval + 1)
+
     def process_offline(self, activations, **kwargs):
         """
         Detect the tempi from the (beat) activations.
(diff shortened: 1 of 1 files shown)


madmom/features/tempo.py
```diff
@@ -847,11 +851,13 @@ def __init__(self, min_bpm=MIN_BPM, max_bpm=MAX_BPM, act_smooth=ACT_SMOOTH,
             hist_smooth=hist_smooth, fps=fps, online=online, **kwargs)
         # save additional variables
         if self.online:
-            self.buffer = BufferProcessor(int(buffer_size * self.fps))
+            self._buffer = BufferProcessor(int(buffer_size * self.fps))
```
superbock: I'd prefer having access to the buffer, thus the leading underscore is not necessary.

madmom/features/tempo.py
```diff
@@ -898,11 +904,19 @@ def online_interval_histogram(self, activation, reset=True):
         # reset to initial state
         if reset:
             self.reset()
+        # select relevant activations from buffer for sutraction
+        buffer = self._buffer.buffer[self.min_interval:self.max_interval + 1]
```
superbock: I'd prefer the variable called activations, because this is basically what it is.

madmom/features/tempo.py
```diff
@@ -898,11 +904,19 @@ def online_interval_histogram(self, activation, reset=True):
         # reset to initial state
         if reset:
             self.reset()
+        # select relevant activations from buffer for sutraction
```
superbock: typo: subtraction

SebastianPoell:
Here you go :)

Issue #306, opened 2017-07-08
replace MIDI code


superbock:
Since we now have online streaming capabilities, it is time to replace the current MIDI code with something capable of MIDI streams. Our current code is based on [Python-midi](https://github.com/vishnubob/python-midi) which hasn't seen an update within almost 3 years and is not compatible with both Python 2 and 3.

[Mido](https://github.com/olemb/mido) seems to be a good choice, since it is actively maintained and used by many other projects. It offers streaming capabilities and is compatible with both Python 2 and 3.

Hopefully a change also resolves #271 and #192.

fdlm:
Will the new code be compatible with the old interface? If not, we should provide a reasonable deprecation timeline.

superbock:
Right now there is no plan to remove the old code too soon, simply because there is no need to.

The new code is more or less compatible to the old one. The only thing I changed is that `notes`, `tempi` and `time_signature` are now properties instead of methods. Changing the unit (i.e. seconds, ticks or beats) can now be done by changing the `MIDIFile.unit` instead of passing the unit to the function.

I think the old way was ok, but didn't like it too much, so it seems to be a good time to think about an API change when refactoring everything. More details about what changed/will change in #308.

Issue #115, opened 2016-03-11
certain evaluation command line arguments are ignored


superbock:
In #11 the whole evaluation was refactored. Since this PR (more specifically: commit 41d61a152357) some command line arguments are ignored, since the `dest` argument is missing in the parsers.

The evaluation itself works fine, but some parameters are not set if called from the `evaluate` script.

What's a good way to test scripts?

Issue #114, opened 2016-03-10
Sound Pressure Level and RMS should be computable on FramedSignals


fdlm:
Right now, it is only possible to compute SPL and RMS on the whole signal. It should also work on `FramedSignal` instances, where the function returns a SPL/RMS value for each frame.

superbock:
Yes, absolutely! We have the following possibilities:

1. change the function to test the type of input
2. add a `spl()` and `rms()` methods to the `FramedSignal`
3. define classes `SPL` and `RMS` classed which could return the right values

I prefer 2. or 3. since it keeps the function to work an simple arrays only.

PR #308, opened 2017-07-10
[WIP] replace old MIDI code in utils.midi with mido

superbock:
This addresses #306. So far the transition was rather simple. However, mido has some unwanted behaviour and bugs which should be fixed upstream rather than including the fixes here.

- MIDI file or message representations are sometimes a bit odd (see mido issue [#101](https://github.com/olemb/mido/issues/101))
- unit conversion contains bugs (see mido issue [#102](https://github.com/olemb/mido/issues/102) and corresponding PR [#114](https://github.com/olemb/mido/pull/114))
- mido supports only seconds or ticks as time units (see mido PR [#115](https://github.com/olemb/mido/pull/115)).

The new code is more or less compatible to the old one, however `notes`, `tempi` and `time_signature` are now properties instead of methods.

TODOs before this PR can be merged:

- [ ] import `midi` into `io` namespace per default? Probably via try/except.
- [ ] replace `process_notes` with a better solution


superbock:
Closing, since these changes are incorporated in #346.

PR #313, opened 2017-07-13
[WIP] Refactor IO functions

superbock:
This is an alternative to #312 which adds sub-packages to `io`.

The first one is `io.audio` which constitutes the functionality of `audio.ffmpeg` and all the loading functions from `audio.signal`.

Another one will be `io.midi`, which is handled in a separate PR (#308).

TODO / questions:
- [ ] `expand_notes()` was moved to `utils`, not sure if this is a good place
- [x] handling of MIREX output formats should maybe go into a separate `io.mirex` module **EDIT:** since it is only a single function which has different output formats, we'll leave it where it is with a distinct name
- [ ] `write_notes_midi` should go into `io.midi` (e.g. rebase #308 on top of this PR before merging)
- [x] check which `load_chords` function to use
- [x] unify loading of events/onsets/beats, see issue #338
- [ ] move `filter_downbeats` to io-module or rewrite as `write_downbeats`?
- [x] refactor tests after the final structure is defined
- [x] move `load_key` to `io`


superbock:
@fdlm, which `load_chords` should be used? [The old one in `madmom.io`](https://github.com/CPJKU/madmom/blob/refactor_io_functions/madmom/io/__init__.py#L336) or [the newer one in `madmom.evaluation.chords`](https://github.com/CPJKU/madmom/blob/refactor_io_functions/madmom/evaluation/chords.py#L289).

I haven't checked if they return the exact same values -- but they use different keys for the return dict, i.e. 'label' vs. 'chord'. Additionally, these files define `CHORD_DTYPE` differently. Should I open a separate issue for this or should we resolve it within this PR? Unfortunately I did not realise this earlier...

fdlm:
The thing is that the  `madmom.features.chords.load_chords` just loads the file without interpreting the chord labels (that's why the column is called 'label'), while `madmom.evaluation.chords.load_chords` computes a numeric representation of the chord label.

I think for now, we can keep `evaluation.load_chords` and ditch the other, and I will adapt some of my other code (outside madmom) that uses the old `load_chords`.

If you want, I can take care of this.

superbock:
That would be great! Another possibility would be to simply include both the labels and the numeric representation in the dictionary. Sounds like these are equivalent.

fdlm:
They are not exactly equivalent, because multiple labels can map to the same representation (e.g. `chord('A:min') == chord('A:(1,b3,5)')`). Currently, I cannot think of a use case for using the string label, that's why I suggested to just use the load function in `madmom.evaluation.chords` and ditch the other (it is trivial to implement if someone needs it).

superbock:
Makes sense. However, there are still conflicting variants of `CHORD_DTYPE`. Which should be used? Or should we rename one of them or both to reflect the integer/string nature of the last column?

fdlm:
I think renaming is the way to go. I'm working on this, see #320.

madmom/evaluation/key.py
```diff
@@ -1,74 +1,7 @@
...
-    """
-    tonic, mode = key_label.split()
-    key_class = _KEY_TO_SEMITONE[tonic.lower()]
-    if mode in ['minor', 'min']:
-        key_class += 12
-    elif mode in ['major', 'maj']:
-        key_class += 0
-    else:
-        raise ValueError('Unknown mode: {}'.format(mode))
-    return key_class
```
superbock: Shouldn't the `key_label_to_class` function stay in the evaluation module? It seems to me that the integer representation is some kind of internal mapping used for evaluation, isn't it?

fdlm: Well, yes, kind of. Maybe it's cleaner to have the mapping in the evaluation module, right!

superbock:
Superseded by #346, closing.

PR #314, opened 2017-07-17
[WIP] Online beat trackers

SebastianPoell:


madmom/features/beats.py
```diff
@@ -554,6 +569,72 @@ def process(self, activations, **kwargs):
...
+        # reset to initial state
+        if reset:
+            self.reset()
+        beats_ = []
+        for activation in activations:
+            # shift buffer and put new activation at end of buffer
+            buffer = self.buffer(activation)
+            # create an online interval histogram
+            histogram = self.tempo_estimator.online_interval_histogram(
+                activation, reset=False)
```
superbock: You should pass the `reset` argument to the `tempo_estimator` as well, i.e. `reset=reset`. Otherwise the beat tracker resets its state, but the tempo estimation still keeps its old state.

madmom/features/beats.py
```diff
@@ -554,6 +569,72 @@ def process(self, activations, **kwargs):
...
+            # get the dominant interval
+            interval = self.tempo_estimator.dominant_interval(histogram)
+            # compute current and next beat times
```
superbock: This comment is a bit misleading, especially the part of next beat. If I get it right it is the first position which is possible to be the next beat, right?

madmom/features/beats.py
```diff
@@ -554,6 +569,72 @@ def process(self, activations, **kwargs):
...
+            # only detect beats again after at least min_interval frames
+            detections = []
+            if cur_beat >= next_beat:
+                detections = detect_beats(buffer, interval, self.look_aside)
+            # if a detection falls within the last few frames and exceeds a
+            # certain threshold a beat was detected. this is important because
+            # for every frame the tempo may change and therefore the last beat
+            # can easily be missed.
+            if len(detections) > 0 and detections[-1] >= len(buffer) - 5 and \
+               buffer[detections[-1]] > self.threshold:
```
superbock: The indentation is weird and the if-statement is very long. I'd prefer to have multiple logical expressions and then test for them, e.g.:
```
cond_1 = detections[-1] >= len(buffer) - 5
cond_2 = buffer[detections[-1]] > self.threshold

if detections and cond_1 and cond_2:
```
Of course, more meaningful names would be nice, but you get the idea. What I still don't like is some hard coded values such as `-5`. Why 5 frames?

madmom/features/beats.py
```diff
@@ -554,6 +569,72 @@ def process(self, activations, **kwargs):
...
+                    self.beat_counter = 5
+                if self.beat_counter > 0:
+                    display.append('| X ')
+                else:
+                    display.append('|   ')
+                self.beat_counter -= 1
+                # display tempo
+                display.append('| %5.1f | ' % float(self.fps * 60 / interval))
+                sys.stderr.write('\r%s' % ''.join(display))
+                sys.stderr.flush()
```
superbock: We should consider refactoring the visualisation stuff as a mixin, which `BeatTrackingProcessor ` and `DBNBeatTrackingProcessor` could inherit from. Most of the stuff here is copy/pasted from `DBNBeatTrackingProcessor`, but not exactly the same way.

Some more notes: we should then use some constants to define how long the beats are displayed (e.g. `self.beat_counter = DISPLAY_DURATION`) or the inertia of the RNN beat activation strength (e.g. `DISPLAY_INERTIA`).

tests/test_features_beats.py
```diff
@@ -79,6 +79,25 @@ def test_process(self):
         self.assertTrue(np.allclose(beats, [0.11, 0.45, 0.79, 1.13, 1.47,
                                             1.81, 2.15, 2.49]))
 
+    def test_process_online(self):
+        processor = BeatTrackingProcessor(fps=sample_blstm_act.fps,
```
superbock: minor note: use `fps=sample_lstm_act.fps` instead.

SebastianPoell:
Thx for all your comments, will work through all of them and then amend ;)

madmom/features/beats.py
SebastianPoell: I just tested different values here and this just seemed to work best. But maybe i can find something better here. For the threshold the intention was that we don't want false positive predictions e.g. on silence.

madmom/features/beats.py
SebastianPoell: I think that would be very nice! Every BeatTracker could show the activation strength, the tempo and whether the current frame is a beat. My only concern is that different trackers may want to show different additional information (like the DBNTracker does with the state space).

madmom/features/beats.py
SebastianPoell: Hmm, actually I think it's ok the way it is. The tempo_estimator gets reset inside the `reset() `method of the BeatTracker and might be called whenever `process_online()` is called.

If I want to compute a whole song then `process_online()` gets called only once with all the activations and `reset=True`. If I would use `reset=reset` here then it would reset the tempo_estimator for each activation while iterating through all activations.

superbock: Ok, I missed the resetting inside `reset()`. Anyways, I still think passing `reset=reset` is the cleaner solution than calling `self.tempo_estimator.reset()`. When iterating over activations frame by frame, reset must be set to false of course -- [and this is exactly what is done in online mode](https://github.com/CPJKU/madmom/blob/master/madmom/processors.py#L759).

SebastianPoell: Ah, I think this might have been a misunderstanding :)

For online mode, the `reset=reset` would be nicer for sure! What I was talking about was actually the single --online mode (sorry for not making this clear). This is where all the activations are passed at once to `process_online()`. And this is why there is a loop over all the activations inside this method. For online mode this loop is always called just once.

Another possibility would be, that for single --online `process_online()` gets called from the outside for each activation. Do you think that makes sense?

madmom/features/beats.py
```diff
@@ -390,6 +390,51 @@ def recursive(position):
     return np.array(positions)
 
 
+def detect_beats_with_ccf(activations, interval, look_aside=0.2):
```
SebastianPoell: I am actually not sure if you want to keep this method :) It is completely interchangeable with the detect_beats() method but seems to produce results which are a little bit worse. On the other hand its much faster. Maybe its an option for the BeatDetector (with constant tempo).

madmom/features/beats.py
SebastianPoell: Ah sorry, I forgot to pass the `online=True` parameter to `process_single()` (I'm calling stuff with a script). Nevermind, works now ;)

madmom/features/beats.py
superbock: The problem you are facing here is the fact that many musical pieces simply don't have a constant tempo.

Although you're allowing some deviation from a constant tempo, this is only considered for each beat individually, i.e. the tempo remains stable and only some jitter is allowed. If one beat is a bit earlier (faster), the position of the next is not based on the current beat, but rather on the previously pre-computed positions.

`detect_beats` on the other hand allows also some deviation from the main tempo, but the position of the next beat is based on this position rather than the pre-computed ones. This makes this method more flexible, but also slower.

If `detect_beats_with_ccf` performs inferior to `detect_beats` on most datasets I don't see a point for including it, even if it is faster. I'd rather speed up `detect_beats` if speed is really an issue.

madmom/features/beats.py
superbock: I was never questioning the rationale behind using a threshold, e.g.`DBNBeatTrackingProcessor` uses it for the very same reason 😏. Back then I added the threshold for the DBN stuff and did not back-port it to the other methods. Maybe we should do so now and add it to the other methods (after evaluating it's positive effect of course).

superbock: Some more notes regarding the threshold:

- What I do is to determine the first/last frame where the activation is above/below the threshold and use only the segment in between. Although this works quite nicely, I am not sure if this is not simply some sort of adaptation towards the annotations...
- The same could be done for multiple segments, not just one per song. Although this sounds logical and straight forward, I am not sure if this segmentation helps in general. We must be aware of the many songs which have segments without any perceivable beats where counting continues. For these cases splitting into segments doesn't make sense at all.

madmom/features/beats.py
SebastianPoell: Yes, i think the same! At least i can use it for my thesis and evaluate it :)

I'm very curious now about the MultiAgent approach because, similar to `detect_beats` it always jumps from beat to beat but considers multiple tempo hypothesis and should be a very good fit for online mode 😋

madmom/features/beats.py
SebastianPoell: Ah ok, cool! Yeah using only "active" segments seems like a good idea. Maybe the distance between those active segments gives a hint whether it should stop tracking or not. 
For online it's a bit different anyway because we don't know how long the break will be so we have to decide on the spot (or maybe with a bit of delay).

superbock:
I merged #292, please rebase your branch on master.

SebastianPoell:
Ok, rebased this branch and applied your stuck-appveyor fixes to the tests. I think the -0.1% Coverage is due to the CCF-Commit (which can be omitted). Anyway, would be awesome if you could take it from here \m/

PR #309, opened 2017-07-11
Chord Evaluation

fdlm:
Adds evaluation functionality for chord sequences.

TODO:
 - [x] Add tests for the evaluation scores.
 - [x] Change `CHANGES.rst`

Comments are welcome.

PR #315, opened 2017-07-24
Minor cosmetic fixes to chord eval code

fdlm:
Adresses final comments in #309.

TODO:
 - [x] What about the `EvaluationMixIn`?

### DIFF
--- a/madmom/evaluation/chords.py
+++ b/madmom/evaluation/chords.py
@@ -30,7 +30,8 @@ References
 """
 
 import numpy as np
-from madmom.evaluation import evaluation_io
+
+from . import evaluation_io
 
 CHORD_DTYPE = [('root', np.int),
                ('bass', np.int),
@@ -208,7 +209,6 @@ def interval_list(intervals_str, given_pitch_classes=None):
     ----------
     intervals_str : str
         List of intervals as comma-separated string (e.g. 'b3, 5').
-
     given_pitch_classes : None or numpy array
         If None, start with empty pitch class array, if numpy array of length
         12, this array will be modified.
@@ -336,7 +336,6 @@ def evaluation_pairs(det_chords, ann_chords):
     ----------
     det_chords : numpy structured array
         Chord detections with 'start' and 'end' fields.
-
     ann_chords : numpy structured array
         Chord annotations with 'start' and 'end' fields.
 
@@ -348,6 +347,7 @@ def evaluation_pairs(det_chords, ann_chords):
         Detected chords of evaluation segments.
     durations : numpy array
         Durations of evaluation segments.
+
     """
     times = np.unique(np.hstack([ann_chords['start'], ann_chords['end'],
                                  det_chords['start'], det_chords['end']]))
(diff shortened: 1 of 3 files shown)


madmom/evaluation/chords.py
```diff
@@ -856,6 +865,7 @@ class ChordSumEvaluation(ChordEvaluation):
         Evaluation objects.
     name : str
```
superbock: str, optional ;)

madmom/evaluation/chords.py
```diff
@@ -892,6 +902,7 @@ class ChordMeanEvaluation(ChordEvaluation):
...
```
superbock: str, optional ;)

madmom/evaluation/chords.py
```diff
@@ -720,7 +729,7 @@ class ChordEvaluation(object):
         ('undersegmentation', 'UnderSegmentation'),
     ]
 
-    def __init__(self, detections, annotations, name, **kwargs):
+    def __init__(self, detections, annotations, name='', **kwargs):
```
superbock: I don't have a strong opinion on using `''` instead of `None`, but in the other two classes you use `None`...

superbock:
Regarding the mixin: I think the `metrics` dict is a nice way of having all metrics accessible in a uniform way.

Why/how do you want to reconsider the mixin? In general or simply for the chord evaluation classes. I don't insist on using it, but what's the problem of inheriting the `metrics` dict -- especially since `METRIC_NAMES` are defined anyways.

madmom/evaluation/chords.py
fdlm: This is because in the other classes, I need to create the default string based on other arguments of the function, which cannot be done directly in the method signature. However, I will change this to `name=None` to make it consistent.

madmom/evaluation/chords.py
fdlm: ✓

madmom/evaluation/chords.py
fdlm: ✓

fdlm:
I think we should reconsider it, because:

 - It is nice to have the `metrics` attribute, but the fact is that we don't use it anywhere, do we? The `tocsv` and `totex` formatters rely on `getattr` and `METRIC_NAMES`.
 - Further, as I wrote earlier, I think `__len__` it ultimately useless, because it can only return integers. What is the length of a single evaluation object, then? Always 1? Then it is useless. The number of "events" evaluated? Then it is very rigid, e.g. we can't use it to describe the length in seconds of evaluated pieces.
 - Finally, every evaluation we have implements their own `tostring` method, so having a generic one is not very usefull.

That said, I think such a mixin might be useful to implement generic `MeanEvaluation` and `SumEvaluation` classes. On the other hand, generic aggregation classes could also rely only on `METRIC_NAMES` and `getattr`, without the need for a mixin. Anyways, the question is if we want to do this at all, because all the evaluations seem to work and I don't see an immediate benefit from refactoring the whole thing. It would definitely be low on my ToDo list. 

To be consistent with the other evaluation classes, I will derive `ChordEvaluation` from `EvaluationMixin`.

superbock:
I think your first and last sentences are enough to simply derive from this mixin (although technically speaking it isn't a mixin any more).

I think we should discuss your other points in a separate issue.

PR #316, opened 2017-07-25
Refactor downbeats and add bar tracking

superbock:
This PR moves all downbeat & pattern tracking functionality into a separate `downbeats` module. It additionally adds bar tracking (i.e. track the downbeats given the beat positions) functionality.

~~Although this PR is more or less ready, there are a couple of things to be discussed:~~

- [x] Should `RNNBarProcessor` return a single 2D numpy array containing the beat times as the first column and the probability to be a downbeat position as the second column (eventually returning NaN as its last (or: first, see next point) value. **EDIT:** returns a 2D array with `NaN` as the last downbeat probability. This also enables us to simply save the bar activations.
- [x] If we return the same number of downbeat probabilities as beat positions, should these probabilities be based on the features of the following beat period (as it is now) or on the one which just passed? Or: should these be aligned differently, e.g. centred? **EDIT:** features are centred around the beat (or interpolated subdivision) and num_subdivision features (for the current) beat are returned. We might change this for online mode, however.
- [x] Should the position of the window to aggregate/summarise the features around the (sub-)beat positions/divisions be coupled with the `origin` parameter of `FramedSignal`? **EDIT:** deferred for later (i.e. when implementing online downbeat trackers).
- [x] Check the impact/usefulness of interpolating missing values when synchronising features.  **EDIT:** removed, since no difference was observed.
- [x] Decide how pattern transitions in `beats_hmm.MultiPatternTransitionModel` should be parametrised, right now the parameters are kind of overlapping. **EDIT:** use `transition_prob` to be able to model transitions between patterns. 
- [x] Fix inability to transition between patterns with more than 1 entry state.
- [x] Decide which classes to import into the `features`-namespace.
- [x] Find and fix memory leak (suspect: `CLPChromaProcessor`). **EDIT:** (at least partially) fixed in #322.


superbock:
On the Ballroom set, interpolation has exactly 0 impact, thus removing it completely.

superbock:
Rebased on master after merging #322 solves the memory issue. `CLPChromaProcessor` makes heavy use of `FramedSignal` which in turn triggered the slice memory/reference leak.

superbock:
`MultiPatternStateSpace` and `MultiPatternTransitionModel` somehow use different arguments than their beat and bar counterparts. Maybe we should change this to be more consistent?

madmom/ml/hmm.pyx
```diff
@@ -99,6 +99,52 @@ class TransitionModel(object):
         return np.log(self.probabilities)
 
     @staticmethod
+    def make_dense(states, pointers, probabilities):
+        """
```
fdlm: This does not relate to this PR directly, but I'm not sure if `make_dense` and `make_sparse` are good names here, because imho they just convert to/from different sparse representations of the transition matrix (CCO to CSR, see https://en.wikipedia.org/wiki/Sparse_matrix)...

madmom/ml/nn/layers.py
```diff
@@ -568,6 +568,16 @@ def __getstate__(self):
         return state
 
     def __setstate__(self, state):
+        # TODO: old models have a 'hid_init' instead of an 'init' attribute
+        #       remove this unpickling code after updating all models
```
fdlm: I think we should track such ToDos in the issue tracker. ToDo items in code tend to be forgotten after some time... :smile:

madmom/ml/hmm.pyx
superbock: Yes, this is something which bothered me as well, especially since it doesn't become dense. How about compress/de-compress instead? We can deprecate `make_sparse` then -- although I don't think a lot of people use it, since this is a rather low-level routine. We could probably even prefix it with _.

madmom/ml/nn/layers.py
superbock: I am not sure if all of these TODOs should go into the issue tracker, since there are plenty and some are simply not too urgent. However I still mark them with TODO, since then at least Pylint lists them and they won't be forgotten totally.

madmom/ml/hmm.pyx
superbock: I will still merge this PR as it is now, since there exists already a `from_dense` method, so this should be addressed jointly. Created issue #335.

PR #320, opened 2017-08-07
Refactor Chord IO Functions

fdlm:
This is a first take on refactoring chord io. However, currently I'm leaning towards the following change to make things more general:

 - In `madmom.io`, do not provide `load_chord` and `write_chord` functions, but a more general `load_segment`, which can load anything that is of the format `<start> <end> <label>`.

I will think more about this in the coming days. This PR is mainly to check if things still work on various Python versions.

### DIFF
--- a/bin/CNNChordRecognition
+++ b/bin/CNNChordRecognition
@@ -11,7 +11,7 @@ import argparse
 
 from madmom.features import (ActivationsProcessor, CNNChordFeatureProcessor,
                              CRFChordRecognitionProcessor, )
-from madmom.io import write_segments
+from madmom.io import write_chords
 from madmom.processors import IOProcessor, io_arguments
 
 
@@ -79,7 +79,7 @@ def main():
     else:
         # load conditional random field
         chord_processor = CRFChordRecognitionProcessor(**vars(args))
-        out_processor = [chord_processor, write_segments]
+        out_processor = [chord_processor, write_chords]
 
     processor = IOProcessor(in_processor, out_processor)
     args.func(processor, **vars(args))
--- a/bin/CNNChordRecognition
+++ b/bin/CNNChordRecognition
@@ -11,7 +11,7 @@ import argparse
 
 from madmom.features import (ActivationsProcessor, CNNChordFeatureProcessor,
                              CRFChordRecognitionProcessor, )
-from madmom.io import write_chords
+from madmom.io import write_segments
 from madmom.processors import IOProcessor, io_arguments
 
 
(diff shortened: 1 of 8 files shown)


superbock:
I think a `load_segments` function is a great idea. Maybe we should keep it rather generic and simply return the last column as a string unless a format for that (or all) column(s) is specified. Then the `load_chords` function could simply specify the format to be `CHORDS_DTYPE` (or whatever).

madmom/io/__init__.py
```diff
@@ -333,62 +345,67 @@ def write_notes_mirex(notes, filename, duration=0.6):
...
+            s, e, l = line.split()
+            start.append(float(s))
+            end.append(float(e))
+            label.append(l)
+
+    segments = np.zeros(len(start), dtype=SEGMENT_DTYPE)
+    segments['start'] = start
+    segments['end'] = end
+    segments['label'] = label
+    return segments
```
superbock: Why not simply `np.loadtxt(filename, dtype=SEGMENT_DTYPE, ndmin=1)`?

Second question: why are you using a single dimension? I would have expected the array to have as many lines as segments, i.e. use `ndmin=2` in the above example.

madmom/evaluation/chords.py
```diff
@@ -46,6 +47,29 @@
...
 
 
+def encode(chord_labels):
+    """
+    Encodes chord labels to numeric interval representations.
+
+    Parameters
+    ----------
+    chord_labels : numpy structured array
+        Chord segments in `madmom.io.CHORD_LABEL_DTYPE` format
```
fdlm: Yes, this is a leftover of refactoring. Thanks!

bin/CNNChordRecognition
```diff
@@ -11,7 +11,7 @@ import argparse
 
 from madmom.features import (ActivationsProcessor, CNNChordFeatureProcessor,
                              CRFChordRecognitionProcessor, )
-from madmom.io import write_chords
+from madmom.io import write_segments
```
fdlm: Yes, I agree!

madmom/io/__init__.py
fdlm: 1) Because it is more than 4 times faster this way. Loading data was the bottleneck.
2) This is how structured arrays work. They only have a single dimension (length). The columns are not a dimension, but handled using the dtype.

superbock: Ok, just asking.

I once also did the loading of events manually, but decided to trade speed for flexibility. E.g. `np.loadtxt` ignores empty lines, can handle basically every possible separator, has nice handling for comments (in case of chords it is perfectly working with `comments='# '`), etc...

Your decision, though.

fdlm: Sure, these are nice things to have. I will probably do some refactoring in the chord evaluation code anyways. I'll run some more performance tests and see whether it still pays off to do it manually.

superbock:
Is this PR still work in progress or can I merge it (into the not-to-be-merged-into-master-yet branch)?

fdlm:
I'll check in the next couple of days and let you know...

fdlm:
Should be fine if tests run through...

PR #322, opened 2017-08-10
prevent memory leak when accessing signal frames

### LINKED ISSUE #321: Memory leak
Somewhere in madmom is a memory leak. It revealed while I was working on #316 and testing different settings of the new `BarTracker`. My first guess was that `CLPChromaProcessor` eats memory, since it appears mostly when using this processor. Further investigation revealed that `FramedSignal` seems to be the culprit, or at least parts of it.

So far I have not been able to really spot the leak, however `gc.get_objects()` listed countless slices objects, all having a start and stop 2048 apart, thus I suspect that the `signal_frame()` function is the problem. But I can't see anything fishy inside, except the last line `return signal[start:stop]`.

If I perform countless numpy array view operations, I can somehow replicate the behaviour:

```
import nump as np
import gc
x = np.arange(100000000)  # uses ~800MB memory
for i in range(len(x) - 20):
    x[i: i+20]  # memory rapidly grows up to ~3.8GB, then stays at this level, goes down to ~3GB at the end of the loop
del x  # memory goes down to 2.3GB
gc.collect()  # remaining memory is freed
```

I expected that `del x` would free the memory, however this is not the case. Is this behaviour only local to my machine or can someone else replicate this?

superbock:
Fixes #321

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,6 +14,7 @@ Bug fixes:
 * Fix double beat detections in `BeatTrackingProcessor` (#298)
 * Fix ffmpeg unicode filename handling (#305)
 * Fix STFT zero padding (#319)
+* Fix memory leak when accessing signal frames (#322)
 
 API relevant changes:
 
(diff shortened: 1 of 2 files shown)


superbock:
Merging, since it limits the memory consumption of `BarTracker` to a stable ~100MB -- before it easily exceeded 2GB in batch mode processing many files.

PR #327, opened 2017-08-22
Quantisation of events and notes

### LINKED ISSUE #325: Quantize Events - Wrong dimensions returned.
I'm trying to read and quantize a midi file.
```
madmom.utils.quantize_events(madmom.utils.midi.MIDIFile.from_file('http:..www.jsbach.net.midi.bwv996.mid.mid').notes(), 100)
```
This returns a one-dimensional list of float64s. The description, as per the documentation, says it returns the events. I'm a bit confused as to the output here.

superbock:

- Quantisation of events does not alter them
- Check for correct dimensionality (i.e. 1D) of input data
- Add functionality to be able to quantise notes, i.e. return a 2D representation of the notes.

This pull request fixes #325.

### DIFF
--- a/madmom/utils/__init__.py
+++ b/madmom/utils/__init__.py
@@ -347,8 +347,11 @@ def combine_events(events, delta, combine='mean'):
     # return immediately if possible
     if len(events) <= 1:
         return events
-    # create working copy
-    events = np.array(events, copy=True)
+    # convert to numpy array or create a copy if needed
+    events = np.array(events, dtype=np.float)
+    # can handle only 1D events
+    if events.ndim > 1:
+        raise ValueError('only 1-dimensional events supported.')
     # set start position
     idx = 0
     # get first event
@@ -397,6 +400,9 @@ def quantize_events(events, fps, length=None, shift=None):
     """
     # convert to numpy array or create a copy if needed
     events = np.array(events, dtype=np.float)
+    # can handle only 1D events
+    if events.ndim != 1:
+        raise ValueError('only 1-dimensional events supported.')
     # shift all events if needed
     if shift:
         events += shift
--- a/madmom/utils/__init__.py
+++ b/madmom/utils/__init__.py
@@ -380,14 +380,14 @@ def quantize_events(events, fps, length=None, shift=None):
 
     Parameters
     ----------
-    events : numpy array
+    events : list or numpy array
         Events to be quantized.
     fps : float
         Quantize with `fps` frames per second.
     length : int, optional
         Length of the returned array.
     shift : float, optional
-        Shift the events by this value before quantisation
+        Shift the events by this value before quantization.
 
     Returns
     -------
@@ -395,8 +395,8 @@ def quantize_events(events, fps, length=None, shift=None):
         Quantized events.
 
     """
-    # convert to numpy array if needed
-    events = np.asarray(events, dtype=np.float)
+    # convert to numpy array or create a copy if needed
+    events = np.array(events, dtype=np.float)
     # shift all events if needed
     if shift:
         events += shift
--- a/madmom/utils/__init__.py
+++ b/madmom/utils/__init__.py
@@ -388,9 +388,10 @@ def quantize_events(events, fps, length=None, shift=None):
     fps : float
         Quantize with `fps` frames per second.
     length : int, optional
-        Length of the returned array.
+        Length of the returned array. If 'None', the length will be set
+        according to the latest event.
     shift : float, optional
-        Shift the events by this value before quantization.
+        Shift the events by `shift` seconds before quantization.
 
     Returns
     -------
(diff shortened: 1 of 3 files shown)


madmom/utils/__init__.py
```diff
@@ -419,6 +426,97 @@ def quantize_events(events, fps, length=None, shift=None):
     return quantized
 
 
+def quantize_notes(notes, fps, length=None, num_notes=None, shift=None,
+                   offset=None, velocity=None):
+    """
+    Quantize the notes with the given resolution.
```
fdlm: I think the description needs to be elaborated. I would expect that "quantizing the notes" would "snap" their start and end points to a grid. This function, however, does something else. I'm not sure how to describe this concisely...

madmom/utils/__init__.py
```diff
@@ -419,6 +426,97 @@ def quantize_events(events, fps, length=None, shift=None):
...
+    Parameters
+    ----------
+    notes : 2D numpy array
+        Notes to be quantized.
+    fps : float
+        Quantize with `fps` frames per second.
+    length : int, optional
+        Length of the returned array. If 'None', the length will be set
+        according to the latest sounding note.
+    num_notes : int, optional
```
fdlm: The name of this parameter is easily confused with the "number of notes in the notes array". Maybe changing the description to something like "Number of possible pitches"? Not sure...

madmom/utils/__init__.py
```diff
@@ -419,6 +426,97 @@ def quantize_events(events, fps, length=None, shift=None):
...
+        Number of notes of the returned array. If 'None', the number of notes
+        will be based on the highest note number in the `notes` array. All note
+        numbers equal or greater this value will not be included in the
+        returned quantized array.
+    shift : float, optional
```
fdlm: I don't think `shift` or `offset` should be part of this function.

1) It is very easy to implement both outside of the function
2) It makes the function signature unnecessarily complex

madmom/utils/__init__.py
superbock: Right, I didn't see it that way. Maybe `expand_notes` would be a more appropriate name?

madmom/utils/__init__.py
superbock: True, at least for `offset`. On the other hand, since `shift` can be a float and rounding is used inside the function, moving this outside the function may introduces more errors.

madmom/utils/__init__.py
superbock: Renamed to `num_pitches`.

madmom/utils/__init__.py
superbock: Added a more descriptive docstring.

madmom/utils/__init__.py
fdlm: IMHO the following is not more error-prone:
```python
notes[:, 0] += 2.3
qnotes = quantize_notes(notes)
```
than this:
```python
qnotes = quantize_notes(notes, shift=2.3)
```
If there is a common use case for the shift, then the second option is fine, but if it is a special case, I vote for the first one, because I think a function should do only one thing, if possible. That said, I don't have a strong opinion on this specific case.

madmom/utils/__init__.py
```diff
@@ -419,6 +426,87 @@ def quantize_events(events, fps, length=None, shift=None):
...
+def quantize_notes(notes, fps, length=None, num_pitches=None, shift=None,
+                   velocity=None):
+    """
+    Quantize the notes with the given resolution.
+
+    First a sparse 2D array with the length and number of pitches is created.
+    If no `length` and `num_pitches` are given the values are inferred from the
+    notes. Then each note is put into the column corresponding to its pitch
+    and all frames between the note's onset and offset (i.e. onset + duration)
+    are set to the velocity of the note.
```
fdlm: Already much better! I would try to describe **what** the function does, and not **how** it does it, maybe like this:

"Create a sparse 2D array whose rows correspond to points in time (according to `fps` and `length`), and columns to note pitches (according to `num_pitches`). The values of the array correspond to the velocity of a pitch at a given point in time (based on the note pitch, onset, duration, and velocity). If no values for `length` and `num_pitches` are given, they are inferred from `notes`."

madmom/utils/__init__.py
```diff
@@ -419,6 +426,87 @@ def quantize_events(events, fps, length=None, shift=None):
...
+    fps : float
+        Quantize with `fps` frames per second.
+    length : int, optional
+        Length of the returned array. If 'None', the length will be set
+        according to the latest sounding note.
+    num_pitches : int, optional
+        Number of pitches of the returned array. If 'None', the number of
+        pitches will be based on the highest pitch in the `notes` array.
+    shift : float, optional
+        Shift the events by `shift` seconds before quantization.
```
fdlm: If we keep the `shift` parameter, I think it should say "Shift the **notes**...

fdlm:
I think there was a comment of yours about renaming of the function to `expand_notes`, or something like that. I can't find it right now. Anyways, I think with a useful description in the docstring, both names should be fine. I would go with whatever is more consistent with the rest of the codebase.

madmom/utils/__init__.py
superbock: You are right, I somehow only considered applying the shift after calling the function... I will remove it (and deprecate the shift in `quantize_events`).

PR #333, opened 2017-09-06
[WIP] Multi Agent Beat Tracker

SebastianPoell:
This PR adds another Beat Tracker which is based on Multiple Agents. It already works for offline and online mode but there are some improvements to do. I think there is still a lot of stuff which should be refactored. Ideas and suggestions are highly welcome :)

- [ ] Refactor or answer TODOs inside code
- [ ] Try to refactor more common things out of offline/online
- [ ] Better comments
- [ ] Tests
- [ ] bin/ Runner
- [ ] Try to find better parameters


madmom/features/beats.py
```diff
@@ -1238,6 +1238,380 @@ def add_arguments(parser, min_bpm=MIN_BPM, max_bpm=MAX_BPM,
...
+    """
+    Beat Tracking via Multiple Agents. Oriented towards the paper "Beat
+    Tracking for multiple applications: A multi-agent system architecture with
+    state recovery" by Lobate Oliveira et al., 2012
+
+    Parameters
+    ----------
+    max_agents : int
+        Max number of agents.
+    tempi : int
```
superbock: To be in line with other beat trackers, this should be `num_tempi`.

Additionally, all these parameters should be available in `__init__`, otherwise they cannot be controlled. If they are optional (i.e. initialised with default values) the should be marked accordingly.

madmom/features/beats.py
```diff
@@ -1238,6 +1238,380 @@ def add_arguments(parser, min_bpm=MIN_BPM, max_bpm=MAX_BPM,
...
+            outer window factor to each side depending on the interval.
+        threshold : float
+            Threshold value for accepting a beat.
+        correction_factor : float
+            Allow agent to adapt to errors.
+        inherit_score_factor : float
+            Child agents will inherit a percentage of their parents.
+
+        """
+        # TODO: Should we use @classmethod to set all values from outside?
```
superbock: @classmethod is probably not the right choice for this. Usually these parameters are just added to `__init__`. 

Since you documented them here (as numpydoc requires the init-params to be documented at class-level), these should be available to `__init__`. Vice versa, the params of init must be documented here as well.

madmom/features/beats.py
```diff
@@ -1238,6 +1238,380 @@ def add_arguments(parser, min_bpm=MIN_BPM, max_bpm=MAX_BPM,
...
+        Number of tempi to be considered.
+    induction_time : float
+        Window length in seconds used for inducting the agents.
+
+    """
+    MAX_AGENTS = 30
+    TEMPI = 3
+    INDUCTION_TIME = 2.
+
+    class Agent(object):
```
superbock: I'd rather make this a class at module level. Although it is only used by `MultiAgentBeatTrackingProcessor` it is IMHO cleaner to move this class one level up. I'd rename it to BeatTrackingAgent(Processor) or something like that, maybe add a leading underscore if it is not usable otherwise.

madmom/features/beats.py
```diff
@@ -1238,6 +1238,380 @@ def add_arguments(parser, min_bpm=MIN_BPM, max_bpm=MAX_BPM,
...
+
+        """
+        # TODO: Should we use @classmethod to set all values from outside?
+        INNER_WINDOW = 5
+        OUTER_WINDOW = 0.4
+        THRESHOLD = 0.05
+        CORRECTION_FACTOR = 0.25
+        INHERIT_SCORE_FACTOR = 0.9
+
+        # used for normalizing the score, set from outside
```
superbock: If these are set from outside, normal class variables are just fine (lowercase without the leading _).

madmom/features/beats.py
```diff
@@ -1238,6 +1238,380 @@ def add_arguments(parser, min_bpm=MIN_BPM, max_bpm=MAX_BPM,
...
+                self.__class__(score=self.score * self.INHERIT_SCORE_FACTOR,
+                               interval=self.interval + error,
+                               prediction=self.prediction + error,
+                               beats=self.beats),
+                # create agent with adjusted tempo and prediction by half
+                self.__class__(score=self.score * self.INHERIT_SCORE_FACTOR,
+                               interval=self.interval + int(error * 0.5),
+                               prediction=self.prediction + int(error * 0.5),
+                               beats=self.beats)
+            ]
```
superbock: This could be done in less lines of code with a simple loop.

Maybe a stupid question, but why is the error always added and not subtracted as well?

madmom/features/beats.py
```diff
@@ -1238,6 +1238,380 @@ def add_arguments(parser, min_bpm=MIN_BPM, max_bpm=MAX_BPM,
...
+            # only return agents which are within tempo range
+            return [agent for agent in agents if
+                    self._MIN_INTERVAL <= agent.interval <= self._MAX_INTERVAL]
+
+        def accept(self, activation, idx):
```
superbock: I don't think this simple logic justifies its own function. I'd rather add the check as an additional condition to the (three) places where this function is called.

madmom/features/beats.py
```diff
@@ -1238,6 +1238,380 @@ def add_arguments(parser, min_bpm=MIN_BPM, max_bpm=MAX_BPM,
...
+                if self.beats[-1] < idx - len(activations):
+                    self.accept(act, self.prediction)
+                # update prediction
+                self.prediction += self.interval
+                # create child agents
+                new_agents = self.fork(error)
+                # penalize agent for not detecting the beat
+                self.score -= (abs(error) / frames) * normalization * act
+                # return new child agents
+                return new_agents
```
superbock: This could probably be refactored a bit by moving common code outside the if/else clause and then decide later on whether new agents need to be created or not.

madmom/features/beats.py
```diff
@@ -1238,6 +1238,380 @@ def add_arguments(parser, min_bpm=MIN_BPM, max_bpm=MAX_BPM,
...
+        # tempo estimator
+        if tempo_estimator is None:
+            # import the TempoEstimation here otherwise we have a loop
+            from .tempo import TempoEstimationProcessor
+            # create default tempo estimator
+            tempo_estimator = TempoEstimationProcessor(fps=fps, online=online,
+                                                       **kwargs)
+        self.tempo_estimator = tempo_estimator
+        self.agents = []
+        # TODO: Not sure if thats nice?
```
superbock: No, it's not 😏 . I think these values should be passed to the agents when instantiating them.

madmom/features/beats.py
```diff
@@ -1238,6 +1238,380 @@ def add_arguments(parser, min_bpm=MIN_BPM, max_bpm=MAX_BPM,
...
+        if reset:
+            self.reset()
+        beats_ = []
+        for activation in activations:
+            # shift buffer and put new activation at end of buffer
+            buffer = self.buffer(activation)
+            # induct agents after induction time has passed
+            if self.counter == self.INDUCTION_TIME * self.fps:
+                # create histogram of induction window
+                histogram = self.tempo_estimator.interval_histogram(buffer)
```
superbock: I am not sure, but shouldn't `activation` be passed to the tempo estimator and not the buffer? In the latter case you compute the histogram for more than the last activation which is probably not what you want.

madmom/features/beats.py
```diff
@@ -1238,6 +1238,380 @@ def add_arguments(parser, min_bpm=MIN_BPM, max_bpm=MAX_BPM,
...
+            # if current frame is considered a beat return it as result
+            if is_beat and self.last_beat < beat_distance:
+                beats_.append(self.counter)
+                self.last_beat = self.counter
+            # increase frame counter
+            self.counter += 1
+        # return beat(s)
+        return np.array(beats_) / self.fps
+
+    def induct_agents(self, activations, interval, start=0):
```
superbock: Probably make start a required parameter. Guess it is less error-prone then.

madmom/features/beats.py
```diff
@@ -1238,6 +1238,380 @@ def add_arguments(parser, min_bpm=MIN_BPM, max_bpm=MAX_BPM,
...
+            Global frame number where agents start.
+
+        """
+        from scipy.signal import argrelextrema
+        # get all maxima within activations window
+        maxima = argrelextrema(activations, np.greater)[0]
+        # if no maxima could be found just use max value
+        if len(maxima) == 0:
+            maxima = np.array([activations.argmax()])
+        # pick N maxima indizes where activation is highest
```
superbock: indices

SebastianPoell: Thx

madmom/features/beats.py
SebastianPoell: Hmm, since all three objects created here have different intervals and predictions the only loop-thingy coming to my mind right now would be something like that:

```
predictions = [self.prediction + error, self.prediction + error, self.prediction + int(error * 0.5)]
intervals = [self.interval, self.interval + error, self.interval + int(error * 0.5)]

for p, i in zip(predictions, intervals):
  self.__class__(prediction=p, interval=i...)
```
Is that would you mean? I also want to test if all three child-agents are really needed (they are used in the paper) to make the results better.

Regarding the error variable: It can also be a negative number (if the event occurred before the prediction), so adding should work ;)

madmom/features/beats.py
SebastianPoell: Ah yeah, thats a good point and actually a larger topic ;)

Well, the concept for both offline and online is the same: Always use the first N seconds to estimate the tempo. Then introduce the agents (for offline at the beginning of the song, for online after those N seconds). From that point in time there is no more tempo estimation going on and the agents are themselves responsible to adapt to the changing speed of the music. That's why in the code the online tempo estimator is not used at all, only the offline tempo estimator is used once after those N seconds (Btw. thats also done in a similar way in all papers I found). That also means that for online the first N seconds don't have agents and therefore cannot produce beat predictions which probably leads to a worse result. For that reason in some papers they just ignore those first N induction seconds when evaluating - but I think that would be unfair to all the other tracking algorithms which are capable of producing predictions from the beginning.

One other problem we are facing here is that the online networks tend to give worse predictions in (approx.) the first 5 seconds of a song. And that is exactly the section used to estimate the base-tempi for the agents. So I already thought about how agents could get their tempo updated during runtime but it's difficult because each agent has its own tempo already from the beginning (since we are using multiple tempo estimations) and updates it on the fly. 

Also connected to that is the handling of abrupt tempo changes. Some solutions I found use a kind of "overwatcher" module which is only looking for sudden changes and makes another induction when necessary. I don't know yet whether we will need something like this.

So yeah, right now for offline I get 90% and for online 80% F-Score for the Ballroom set (didn't play much with the parameters yet). If we get some more ideas I will try them out ;)

superbock: Yes, tempo estimation at the very beginning is indeed a problem. Relying on those estimations is thus probably not a very good idea. How about simply starting a fixed number of agents at the beginning of the song and initialise them with tempi distributed uniformly on a logarithmic tempo scale? Soon afterwards the "bad" agents will be killed/die anyways, right?

For example for downbeat tracking I limit the number of modelled tempi to 60 in order to prevent an excessively bloated state space. 60 tempi is enough to cover the whole tempo range from 40..250bpm. Please note, that the downbeat tracker is not able to use any other tempi.

I guess the agents are able to adapt their tempo quite quickly, so maybe this is a good workaround/extension and would render the need for tempo estimation completely.

madmom/features/beats.py
superbock: Yes, something like this, but you could even simplify it to:

```
prediction_errors = [1, 1, 0.5]
interval_errors = [0, 1, 0.5]

for p, i in zip(prediction_errors, interval_errors):
    self.__class__(prediction=self.prediction + int(p * error),
                   interval=self.interval + int(i * error), ...)
```

SebastianPoell: Nice, thx!

madmom/features/beats.py
SebastianPoell: Ah, yeah interesting idea! I will test that (maybe also for offline) ;)

PR #342, opened 2017-10-12
[WIP] Unify load functions

### LINKED ISSUE #338: Unify behaviour of load functions.
Currently, load functions (`load_beats`, `load_chords`, ...) do not all accept the same type of parameters. For the sake of consistency, we should unify this. This will also make them easier to use.

We need to define what each `load_*` function accepts, while keeping in mind that the `*Evaluation` classes should be usable with files, file handles, and directly passed detections.

Here is my suggestion:

 - `load_*` functions *only* accept files or file handles. They are called "load" functions for a reason.
 - Checking whether something is a file (and thus `load_*` needs to be called) or a detection instance should be done within the `__init__` methods of the evaluation classes.

Ping @superbock :smile:

superbock:

All `load_*` functions in `madmom.io` can handle only files or file handles. Functionality to read annotations/detections was moved to `bin/evaluate` accordingly.

This pull request fixes #338.

- [x] adapt alignment evaluation + add sample annotations/detections or remove functionality
- [x] rebase onto master after #313 is merged (or merge into `refactor_io_functions`)


superbock:
@fdlm can you provide alignment annotations/detections to write a meaningful test for alignment evaluation?

fdlm:
I could, by right now this is not on my priority list, since I don't work and do not plan to work on alignment anymore.

superbock:
Providing some samples is absolutely fine. I'd just add a simple test that makes sure that the evaluation metrics don't change.

superbock:
On the other hand, should we remove the alignment evaluation code completely? At least there is nor corresponding `features` module.

fdlm:
I think we can remove it for the reasons you outlined.

PR #345, opened 2017-12-01
Implemented key recognition

fdlm:

This PR adds key recognition functionality to madmom (`features.key.*`, and the `bin/KeyRecognition` program). Additionally, it implements a padding layer and a function layer in the neural network module.

Comments welcome! I will do the tests once the overall structure of this PR is final. 

- [x] Write tests for the new layers.
- [x] Write tests for `features.key`.
- [x] Write tests for `bin/KeyRecognition`.
- [x] Remove single net option, add issue to handle this on a more general level.


fdlm:
By the way, I'm not very happy with the need for a `FunctionLayer`, but this is the only way to go right now if we do not want to implement a special layer for every numpy operation out there.

superbock:
A simple fix would be to simply call the layers in `NeuralNetwork` instead of calling their `activate()` method (https://github.com/CPJKU/madmom/blob/master/madmom/ml/nn/__init__.py#L101). Then a simple partial would do it.

madmom/ml/nn/layers.py
```diff
@@ -945,3 +945,87 @@ def activate(self, data, **kwargs):
...
+        """
+        shape = list(data.shape)
+        data_idxs = [slice(None) for _ in range(len(shape))]
+        for a in self.axes:
+            shape[a] += self.width * 2
+            data_idxs[a] = slice(self.width, -self.width)
+        data_padded = np.empty(tuple(shape))
+        data_padded[:] = self.value
+        data_padded[data_idxs] = data
+        return data_padded
```
superbock: First of all, I'd like to see the NN changes split into a separate commit.
The layer itself looks good to me, however it would be great to unify [with the padding applied for the CNNOnsetProcessor](https://github.com/CPJKU/madmom/blob/master/madmom/features/onsets.py#L802) which needs repeating of the first/last frame N times.

madmom/features/key.py
```diff
@@ -0,0 +1,144 @@
...
+    <madmom.features.key.CNNKeyRecognitionProcessor object at 0x...>
+    >>> proc('tests/data/audio/sample.wav')  # doctest: +ELLIPSIS
+    array([[ 0.00096,  0.00226,  0.0169 ,  0.00138,  0.01767,  0.00352, ...,
+             0.17846,  0.12905,  0.06041,  0.08945,  0.00312,  0.08681]], dtype=float32)
+             
+    """
+ 
+    SINGLE_NET = False
+
+    def __init__(self, nn_files=None, single_net=SINGLE_NET, **kwargs):
```
superbock: I don't like the `single_net` parameter, maybe a more generic `num_models` to state the maximum numbers of models to be evaluated would be more flexible. This could be moved to `NeuralNetworkEnsemble`, then all other programs/algorithms would benefit from it.

We should consider using fewer models in general in order to keep the package smaller. Supplying multiple networks is often not really needed for good performance and are often only supplied to experiment setups. But maybe this discussion should be moved somewhere else.

madmom/ml/nn/layers.py
```diff
@@ -945,3 +945,87 @@ def activate(self, data, **kwargs):
...
+        Indices of axes to be padded
+    value : float
+        Value to be used for padding.
+
+    """
+
+    def __init__(self, width, axes, value=0.):
+        self.width = width
+        self.axes = axes
+        self.value = value
```
superbock: Maybe we should allow literals like 'reflect', 'repeat' or add a separate `mode` like [SciPy does for convolution](https://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.ndimage.filters.convolve.html) in order to be more flexible in the way how to pad the data (see below).

madmom/features/key.py
fdlm: Yes, but wouldn't it be weird to pass a long list of models to the `NeuralNetworkEnsemble`, and also pass a parameter that sets how many to select? You could just pass the shorter list to the ensemble in the first place.

I agree with the number of models. In this case, however, using the ensemble gives +3pp for Pop/Rock music, so I thought it's worth it.

madmom/ml/nn/layers.py
fdlm: That would be nice, but I'd rather implement this when it is needed. Judging from the current literature, everyone uses 0-padding in NN, because it does not make a difference.

madmom/ml/nn/layers.py
fdlm: Hm, ok, then repeating is necessary, after all :)

Do you mean a separate commit or a separate PR? Anyways, I agree that the `PadLayer` could be more flexible, but I still think we can merge it with a bare minimum of functionality and add more afterwards.

fdlm:
How about I do an extra PR for the following NN changes:
 - Call the layers instead of using `.activate()`. This would allow to remove the `FunctionLayer` and thus reduce code complexity.
 - Add a simple `PadLayer` that can be extended afterwards.

madmom/features/key.py
superbock: What I basically wanted to say that this option should go to a more central place if there's a need for it. Following your argumentation, `single_net` is not the way to go then ;)

Regarding the number of models: ensembles usually give a performance boost, that's why I also check in multiple models. I'm fine with 4 models to be checked in into madmom/models, so I think we can defer this discussion and aim for a global solution since this doesn't affect only key estimation.

madmom/ml/nn/layers.py
superbock: I find it weird that the commit message of adding a PaddingLayer is "implemented key recognition feature", thus I think this should be a separate commit. There's no need for a separate PR and we can move the padding of `CNNOnsetDetector` later on, but we should rather think about argument names and needed functionality now than later.

superbock:
Having a separate PR sounds like the way to go then. Feel free to start working on it :)

fdlm:
Ok, the problem is that the `NeuralNetwork` class calls each layer with `reset=True`. Now, if I want to have e.g. `np.reshape` as layer, it will blow up, because it does not accept a `reset` argument. A solution would be to remove the `reset` argument from the signature of `activate`, and let the `NeuralNetwork` do something like `if hasattr(layer, 'reset'): layer.reset()` before calling each layer. However, this would break existing models, I guess. Any ideas?

superbock:
Good point. I am wondering how many functions need to be wrapped. If it is only reshaping and flatten and probably a few others maybe explicitly defining these layers (e.g. `ReshapeLayer`) is ok as well. At least it would be obvious then what these layers do.

fdlm:
For the key model, I need a transpose, reshape, and average layer. The code would be exactly the same, except for the function called on `activate()`.

superbock:
Sounds fine if we have them with dedicated names. I need a (global) averaging layer as well.

fdlm:
Ok, created a separate commit for the layers and the feature. The docs of the layers follow the docs of the respective NumPy functions. Let me know if there's anything more missing. Otherwise, I will start writing the tests.

fdlm:
This is weird - appveyor does not seem to get the correct `models` submodule. This is the error it throws:

```
======================================================================
ERROR: test_process (tests.test_features_key.TestCNNKeyRecognitionProcessorClass)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\projects\madmom\tests\test_features_key.py", line 49, in setUp
    self.processor = CNNKeyRecognitionProcessor()
  File "C:\projects\madmom\madmom\features\key.py", line 96, in __init__
    from ..models import KEY_CNN
ImportError: cannot import name KEY_CNN
----------------------------------------------------------------------
```

Any ideas on how to fix this (otherwise, I guess, the PR is ready to merge).

superbock:
Seems like appveyor is not able to check out the correct revision. I'll investigate this and merge then. Any objections to squashing some of the commits (e.g. PEP8 fixes or nn layer fixes) if I need to rebase it?

fdlm:
No objections, do whatever you deem necessary!

superbock:
Nice work! Squashed some commits, fixed some docstrings and added key recognition to docs. Will merge as soon as the tests pass (appveyor seems to be happy after the models have been merged).

PR #352, opened 2018-01-12
MIDI file warning suppression

Reichenbachian:

Allow error suppression for
- notes
- from_stream
- from_file

This pull request fixes my annoyance when there are hundreds of warnings scrolling down my screen.

### DIFF
--- a/madmom/utils/midi.py
+++ b/madmom/utils/midi.py
@@ -1740,7 +1740,7 @@ class MIDIFile(object):
         midi_file.close()
 
     @classmethod
-    def from_file(cls, midi_file):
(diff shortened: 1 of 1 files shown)


superbock:
Unfortunately, this is not the right way to accomplish what you're looking for (suppress warnings generated by some functions). The suggested way is to handle this in your code, e.g. by catching these warnings. For details, please see: https://docs.python.org/2/library/warnings.html#temporarily-suppressing-warnings

[madmom has a decorator which does exactly that](https://github.com/CPJKU/madmom/blob/master/madmom/utils/__init__.py#L20)

PR #353, opened 2018-01-13
Numpy 1.14 fixes

superbock:

This PR addresses issues when madmom is used with recently released numpy 1.14.

Many errors arose from the fact that `np.savetxt` is not working as before, probably because it closes open files differently. Thus `np.savetxt` was replaced with a pure-Python solution.

Numpy 1.14 changed the default array printing style, thus all doctests are adapted.

The correct dtype for comparison in `issubdtype` is used to avoid future incompatibilities.


tests/test_bin.py
```diff
@@ -917,6 +918,8 @@ def test_txt(self):
         # save activations as txt file
         run_save(self.bin, sample_file, tmp_act, args=['--sep', ' '])
         act = Activations(tmp_act, sep=' ', fps=200)
+        print(len(act), act)
```
fdlm: Are these prints on purpose?

tests/test_bin.py
```diff
@@ -1021,9 +1024,10 @@ def test_mirex(self):
     def test_all_tempi(self):
         run_single(self.bin, sample_file, tmp_result, args=['--all'])
         result = np.loadtxt(tmp_result)
+        print(result)
```
fdlm: Is this print on purpose?

tests/test_bin.py
superbock: no, will remove them.

PR #363, opened 2018-04-17
use pyfftw for FFT

superbock:

Use pyfftw (if available) to compute FFT. Gives up to 40% faster FFT computation.


superbock:
I think it is save to merge, since Travis uses the new pyfftw code, whereas Appveyor doesn't (there were some problems installing pyfftw on Windows). So we have both code paths covered. Probably there's a better way to test both, but I think this suffices.

Issue #378, opened 2018-08-21
Broken online tempo estimation via DBN


SebastianPoell:
### Expected behaviour

Estimating the online tempo via DBN should work for any given activation length.

### Actual behaviour

Hi there, it seems like there is a problem when feeding multiple activations at once to the DBN online_processor:

`ValueError: could not broadcast input array from shape (3015,127) into shape (1000,127)`

After further investigation I discovered that this is because the BufferProcessor cannot handle the case when more data than the original buffer size is fed to the buffer at once:

`buffer = BufferProcessor((1000, 127))`

This works:
`buffer.buffer(np.zeros(shape=(500, 127)))`

This doesn't:
`buffer.buffer(np.zeros(shape=(2000, 127)))`

However, this is not a problem for ACF and COMB, because there, the activations are looped and fed into the buffer one by one.

I think, the best solution would be to adapt the BufferProcessor in such a way that it always cuts the input to a maximum of the given buffer size. What do you think? Can do a PR if you like ;)

### Steps needed to reproduce the behaviour

Just use method=dbn for the tempo detector, set it to online mode and use activations longer than 10s.

SebastianPoell:
Hey, maybe a simpler example. I would expect this to work:

```
 buffer = BufferProcessor(5)
 buffer.buffer(np.array([1,2,3,4,5,6,7,8]))
```

I would expect that after those lines the buffer contains `[4,5,6,7,8]` but instead an error is thrown ;)
Everytime we want to put more numbers into the buffer than the original size, this leads to an error.
This is also the reason the DBN Online-Tempo-Estimator is breaking...

superbock:
I meant that if the buffer is initialised with size 500 and called with 3015 frames, it is expected *not* to work, only if the size is *larger* than the number of expected frames.

Anyways, this is more a design question what should happen if a `BufferProcessor` is called with larger chunks than expected. Right now it throws an error, but this should not be a problem for tempo estimation with the DBN method.

Can you paste a self-contained example, because I am not able to reproduce this problem. What I tried is:

`TempoDetector --method dbn online 10_minutes_audio.wav`

superbock:
Yes, that explains it.

[...]

PR #379, opened 2018-08-21
Fixed error for buffer processor where data bigger than the buffer si…

SebastianPoell:
A fix for #378...


SebastianPoell:
Sounds good :)

PR #396, opened 2018-11-30
Key evaluation updates

### LINKED ISSUE #395: Proposed updates to key evaluation
### Current behaviour

When an estimated key is not the same as an annotated key, there are different ways of evaluating how "far off" the estimated key is.

Currently, in madmom.evaluation.key.error_type the categories outlined in the MIREX competition are used: 'fifth', 'relative', 'parallel' and 'other'. 

Also, directly associated to the type of error, a score is given based on the MIREX key estimation guidelines.

### Desired behaviour

1) It would be nice to enable experimenting with other sets of scores for the different error categories; right now they are strongly coupled in the code and it would be nice to dissociate those.
1) It would be nice to separate key detection errors where the detected key is the relative key of the fifth of the annotated key. For example, if the annotated key is a C major and the estimated key is an E minor, there is only one note that is different between the two scales (F#), just as in G major (which is the fifth of C major). It would then seem logical to provide its own error type to that case as it is still a closely related key compared to the annotated key.

I have submitted a PR related to that issue.

bscherrer:
- separate identification of key error category and score computation: error_scores are now a class variable of the KeyEvaluation class

- addition of a new type of error category: relative_of_fifth that plays well with the strict_fifth option

- additional documentation for the error_type function

- updates to the key evaluation test code and data files to account for the separation of the error category determination and the score to attribute to a given key error

This pull request fixes #395 .


bscherrer:
Could someone have a look at this PR ? It does preserve previous behaviour and adds functionalities that can be useful to other researchers interested in musical key detection evaluation.

Thanks !

madmom/evaluation/key.py
```diff
@@ -80,36 +156,69 @@ def error_type(det_key, ann_key, strict_fifth=False):
...
+    Examples
+    --------
+    >>> from madmom.evaluation.key import error_type
+
+    # annotated: 'C major' / detected: 'C major'
+    >>> error_type(0, 0)
+    'correct'
+
+    # annotated: 'C major' / detected: 'G major': +7 semitones
+    >>> error_type(7,0)
```
fdlm: Whitespace after `7,`:

```
     >>> error_type(7, 0)
```

madmom/evaluation/key.py
```diff
@@ -80,36 +156,69 @@ def error_type(det_key, ann_key, strict_fifth=False):
...
+    'fifth'
+
+    # annotated: 'C major' / detected: 'F major': -7 semitones (modulo 12)
+    >>> error_type(5,0)
```
fdlm: Whitespace after `5,`:
```
     >>> error_type(7, 0)
```

madmom/evaluation/key.py
```diff
@@ -134,14 +243,27 @@ class KeyEvaluation(EvaluationMixin):
...
         self.detection = key_label_to_class(detection)
         self.annotation = key_label_to_class(annotation)
-        self.score, self.error_category = error_type(
-            self.detection, self.annotation, strict_fifth
-        )
+        self.error_category = error_type(self.detection,
+                                         self.annotation,
+                                         strict_fifth,
+                                         relative_of_fifth)
+        self.score = self.error_scores.get(self.error_category, 0.0)
```
fdlm: It would be better to use indexing here so that this line throws an error if an error category is unknown.

madmom/evaluation/key.py
```diff
@@ -184,20 +307,45 @@ def __init__(self, eval_objects, name=None):
...
+            else:
+                indScores += 1
+        return all_the_same
 
     def tostring(self, **kwargs):
         return ('{}\n  Weighted: {:.3f}  Correct: {:.3f}  Fifth: {:.3f}  '
-                'Relative: {:.3f}  Parallel: {:.3f}  Other: {:.3f}'.format(
-                    self.name, self.weighted, self.correct, self.fifth,
-                    self.relative, self.parallel, self.other))
+                'Relative: {:.3f}  Relative of Fifth: {:.3f} '
```
fdlm: We should only report "Relative of Fifth" here if it is actually present in all the evaluation objects.

madmom/evaluation/key.py
```diff
@@ -134,14 +243,27 @@ class KeyEvaluation(EvaluationMixin):
...
+                    'relative': 0.3,
+                    'relative_of_fifth': 0.0,
+                    'parallel': 0.2,
+                    'other': 0.0}
+
+    def __init__(self, detection,
+                 annotation,
+                 strict_fifth=False,
+                 name=None,
+                 relative_of_fifth=False,
```
fdlm: Please add the `relative_of_fifth` parameter next to the docstring of the class.

madmom/evaluation/key.py
bscherrer: done.

madmom/evaluation/key.py
bscherrer: done.

madmom/evaluation/key.py
bscherrer: oops, yes. done.

superbock:
What's the status of this PR? I'm preparing v0.17 and wonder if this is ready to be included.

bscherrer:
I guess I have missed the v0.17 release boat, sorry ... it should be done today (Montreal time zone).

superbock:
No hurry, there's still some more time until the release will be ready. I was just wondering abot the current status.

bscherrer:
OK, cool. I'll still try to aim for finishing this today :)

Bertrand Scherrer, Ph.D.
Audio Research Scientist

LANDR Audio Inc.
160 St Viateur East, Suite 809
Montreal, QC, Canada, H2T 1A8
www.landr.com

madmom/evaluation/key.py
bscherrer: OK, done

madmom/evaluation/key.py
bscherrer: OK

bscherrer:
@superbock and @fdlm all should be good now.

madmom/evaluation/key.py
```diff
@@ -175,29 +301,85 @@ class KeyMeanEvaluation(EvaluationMixin):
...
+        ret += 'Relative: {:.3f}'.format(self.relative) + spacing
+        if self.relative_of_fifth_present:
+            ret += 'Relative of fifth: {:.3f}'.format(self.relative_of_fifth) \
+                   + spacing
+        ret += 'Parallel: {:.3f}'.format(self.parallel) + spacing
+        ret += 'Other: {:.3f}'.format(self.other)
+        return ret
+
+
+def check_key_eval_objects(key_eval_objects):
```
fdlm: I feel this function and the ones below can be simplified a lot by doing something like this (not tested!):
```python
def _check_key_eval_objects(key_eval_objects):
    e = key_eval_objects[0]
    for ke in key_eval_objects[1:]:
        if (ke.error_score != e.error_scores or
            ke.strict_fifth != e.strict_fifth or
            ke.relative_of_fifth != e.relative_of_fifth):
            return False	
    return True
```
Can you check if this is equivalent and if yes, use the simpler code?
Also, I do not think this function makes sense outside of this module, so we should mark it as "private" by using the underscore.

madmom/evaluation/key.py
```diff
@@ -175,29 +301,85 @@ class KeyMeanEvaluation(EvaluationMixin):
...
         ('weighted', 'Weighted'),
     ]
 
     def __init__(self, eval_objects, name=None):
         self.name = name or 'mean for {:d} files'.format(len(eval_objects))
-
+        self.relative_of_fifth_present = False
+        if check_key_eval_objects(eval_objects):
+            self._count_evaluations(eval_objects)
+            if self.relative_of_fifth > 0:
```
fdlm: Shouldn't `self.relative_of_fifth_present` depend only on whether the `eval_objects` compute it or not (i.e., after `check_key_eval_objects`, if `eval_objects[0].relative_of_fifth == True`)?
`self.relative_of_fifth` could be 0 because this error type did not occur, and in this case we want to report it. We only don't want to report it if the `eval_objects` did not compute it in the first place.

superbock:
Ok, v0.17 was delayed a bit ;), but I'm currently working on getting a new version out, so I'd like to check back on this PR and its current status. @bscherrer can you comment on the two comments above? Would be really nice to have this in!

madmom/evaluation/key.py
bscherrer: Indeed, this simpler version works fine. Thanks !

madmom/evaluation/key.py
bscherrer: I think I addressed this comment with the latest commit.

fdlm (approved):
Looking good, thanks!

PR #400, opened 2018-12-07
Level volume when loading ReplayGain-tagged files

### LINKED ISSUE #399: Support loading ReplayGain tagged files
It'd be great if the FFmpeg audio loading helpers supported [ReplayGain](https://en.wikipedia.org/wiki/ReplayGain) tags in the audiofiles.

Many people have their music tagged with this to allow volume levelling, but currently AIUI the FFmpeg `io` helpers (based around `_ffmpeg_call`) don't process these tags. Consequently it's hard to set thresholds and other parameters for feature analysis / display etc with a >12dB average volume range across files.

Luckily FFMpeg has [in-depth support for RG volume adjustment](https://ffmpeg.org/ffmpeg-all.html#volume)...

declension:

 * Use ffmpeg's `volume` filter, and the RG commands within that to sensibly adjust the SPL of audio being loaded when loading files with `io.audio` helpers for this.
 * Provide sensible defaults, and document through `io.audio` chain of helpers.
 * Unit tests (with new RG-tagged FLAC files, other formats should work fine too),
   proving the track / album selection and pre-amp features are working.

This pull request fixes #399.


declension:
@superbock Codeclimate wants me to refactor to avoid the additional (optional) parameters to all those FFMPeg methods, but this becomes a much bigger (and separate, really) change then.

Any preferences here?

superbock:
Ignore the issues. These checks are in place to get hinted if there are some obvious issues. But if the solution would require some serious refactoring for no obvious reason (as in this case), it makes no sense to "fix" them.

One question: I never used the preamp gain of replaygain. Is it of any real use? And if it just amplifies the signal after RG normalisation, it could be omitted since there's already a 'normal' gain parameter in place.

declension:
@superbock did that sound OK, or would you prefer not to have the pre-amp gain at all?

superbock:
No, that sounds perfectly fine. I simply did not have time to review this PR yet, sorry.

madmom/io/audio.py
```diff
@@ -96,6 +97,10 @@ def _ffmpeg_call(infile, output, fmt='f32le', sample_rate=None, num_channels=1,
         Maximum length in seconds to decode.
     cmd : {'ffmpeg','avconv'}, optional
         Decoding command (defaults to ffmpeg, alternatively supports avconv).
+    rg_mode : {'track','album', None}, optional
```
superbock: Default parameters should be listed first.
```suggestion
    replaygain_mode : {'None', 'track','album'}, optional
```

madmom/io/audio.py
```diff
@@ -96,6 +97,10 @@ def _ffmpeg_call(infile, output, fmt='f32le', sample_rate=None, num_channels=1,
...
+        Specify the ReplayGain volume-levelling mode (None to disable).
+    rg_preamp_db : float, optional
+        Increase the volume by this many dB after applying ReplayGain tags.
```
superbock:
```suggestion
    replaygain_preamp : float, optional
        ReplayGain preamp volume change level (in dB).
```

superbock:
Oh, while you're at it, you could add the following to CHANGES.rst:

New features:

* Volume changes according to `ReplayGain` tags can be applied (#400)

madmom/io/audio.py
declension: Oh OK didn't realise that; I haven't used this documentation style really - more used to reST style (or increasingly just Python 3.6+ type annotations a simple docstring). I'll update it - thanks

declension:
@superbock thanks for the review. I think I've addressed everything, and added the same boilerplate for the decode (to disk) function I missed off (plus some new tests).

Not sure why all the 32-bit Windows builds are consistently dying in AppVeyor though -- 
`ImportError('DLL load failed: %1 is not a valid Win32 application.',), <traceback object at 0x0495B148>))`

declension:
Yep, I'd agree that running complex audio feature extraction on 32-bit Windows must be fairly niche by now :slightly_smiling_face:

superbock:
It seems that a newer version of `icc_rt` (2019.0.0-h0cc432a_1) fixed the build failures.

PR #408, opened 2019-01-10
Downbeats: allow weighting of beats per bar

### LINKED ISSUE #402: Downbeat tracking: support weighting for time signatures
### Use case
When I'm running `DBNDownBeatTrackingProcessor`, I'm, passing parameters `beats_per_bar=[3, 4, 6]` to detect from a mixed bunch over many genres (but mostly pop / rock / soul / dance music).

The vast majority is in common (4) time, with a few 6/8 plus a few "difficult" ones. However, by including especially `6` in this list, I find the processor occasionally misidentifying 4/4 pieces, especially rock (and heavier) as having six beats per downbeat in the output.

### Workarounds attempted

* I've tried fiddling with max / min BPMs - not much help
* I've tried altering the number of tempi e.g. `[40, 100, 40]` in the hope that this would weight the probability of, but this felt like it was perhaps harming the downbeat detection overall - is that right? Which parameters need models to be retrained etc?

### Suggestion
Would it be possible to provide weightings or other forms of influencing from the selection of beat per bar given in an array - e.g. just `weights=[1,10,2]` or similar?

Thanks

declension:

 * In the `DBNDownBeatTrackingProcessor`, add an optional weighting array parameter, implicitly defaulting to ones
 * Clean up the handling of lengths into the constructor, it was getting verbose
 * Check weights don't sum to zero, to avoid divide-by-zero pain.
 * Weight the HMM results in log space by normalised weight values, as suggested by @Superbock
 * Add new test to prove that (sufficient, but arbitrary) weighting to 3-time (over 4-time)
   does indeed return 3-time beats results.

This fixes #402.

### DIFF
--- a/madmom/features/downbeats.py
+++ b/madmom/features/downbeats.py
@@ -153,6 +153,11 @@ class DBNDownBeatTrackingProcessor(Processor):
         (down-)beat activation function).
     fps : float, optional
         Frames per second.
+    beats_per_bar_weights : list, optional
+        Weight the beats_per_bar list when choosing.
+        Higher numbers favour the beat number at the same index, e.g.
+        for beats_per_bar of [3, 4], a value here for [1, 2] will bias the
+        the choice towards 4 beats per bar.
 
     References
     ----------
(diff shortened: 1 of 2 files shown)


declension:
Hi @superbock any thoughts on this PR?

superbock:
Sorry, still had no time to review this yet. At a first glance it looks good! I don't like `beats_per_bar_weights` too much, but this is basically what it is ;) I'll look into it more thoroughly soon.

declension:
Cool, thanks!

> I don't like beats_per_bar_weights too much, but this is basically what it is ;)

Agreed, I tried for ages to think of something better, but nothing came up... :thinking:

PR #415, opened 2019-02-14
Allow reading audio from file objects created in memory 

jpauwels:

These changes allow to read audio from file-like objects created in memory (using BytesIO). The first commit contains the actual changes and the second commit implements a unit test.

The original code was able to read audio from physical file handles, but did so in a way that claimed resource ownership instead of leaving it with the calling code (meaning that if you passed a file handle on to madmom I/O, it got closed afterwards, whether you like it or not and there's nothing you can do about it). A side effect of my implementation is that it got rid of this (in my eyes undesirable) behaviour, but then I noticed that a unit test case was written to explicitly verify this behaviour. So I added a third commit to restore the resource management to its original behaviour, but I would advise you to reconsider it (and change the corresponding unit tests).

### DIFF
--- a/madmom/io/audio.py
+++ b/madmom/io/audio.py
@@ -14,6 +14,7 @@ import os
 import subprocess
 import sys
 import tempfile
+import io
 
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


madmom/io/audio.py
```diff
@@ -312,7 +315,7 @@ def decode_to_pipe(infile, fmt='f32le', sample_rate=None, num_channels=1,
                         replaygain_mode=replaygain_mode,
                         replaygain_preamp=replaygain_preamp)
     # redirect stdout to a pipe and buffer as requested
-    if isinstance(infile, Signal):
+    if isinstance(infile, Signal) or isinstance(infile, io.IOBase):
```
superbock: This could be unified to `isinstance(infile, (Signal, io.IOBase)):`

PR #418, opened 2019-02-20
Allow reading audio from file objects created in memory (revised)

superbock:

Updated version of #415, see that PR for details.

This pull request fixes #415.

### DIFF
--- a/madmom/io/audio.py
+++ b/madmom/io/audio.py
@@ -14,6 +14,7 @@ import os
 import subprocess
 import sys
 import tempfile
+import io
 
 import numpy as np
 
@@ -131,6 +132,8 @@ def _ffmpeg_call(infile, output, fmt='f32le', sample_rate=None, num_channels=1,
         in_ac = str(int(infile.num_channels))
         in_ar = str(int(infile.sample_rate))
         infile = str("pipe:0")
+    elif isinstance(infile, io.IOBase):
+        infile = "-"
     else:
         infile = str(infile)
     # general options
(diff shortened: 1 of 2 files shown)


jpauwels:
Thanks for finding the time to do this, you beat me to it :-) In case you're wondering, I have a use case where I'm trying to read audio from the stdin, so this should allow me to skip writing the stdin to a file  just to read it straight after.

madmom/io/audio.py
```diff
@@ -131,6 +132,8 @@ def _ffmpeg_call(infile, output, fmt='f32le', sample_rate=None, num_channels=1,
         in_ac = str(int(infile.num_channels))
         in_ar = str(int(infile.sample_rate))
         infile = str("pipe:0")
+    elif isinstance(infile, io.IOBase):
```
superbock: @jpauwels: This change breaks using any non-wave file in `single` mode, since this mode opens the file as a `io.BufferedReader` which is a subclass of `io.IOBase`. Unfortunately we did not have a test for this, so this was not caught by any test. I opened an issue for this #447.

Since you did not add any test for your use case, I am not sure which specific type `infile` is in your case. Could you please test it and see if it is different to `io.BufferedReader`, than the fix would be quite easy. If not, we have to figure out another way how to distinguish these two.

It would be best if you could add a test for your use case.

jpauwels: Ah, sorry for that. For my use case, I'm hoping to support any format that ffmpeg can read. Up to now I've only tried wave, FLAC and mp3, but it seems better that we'd fix it for all formats, otherwise someone else might run into trouble. I'll try to have a look at the code, but it might not be this week.

PR #421, opened 2019-03-06
[WIP] refine tempo

### LINKED ISSUE #416: Accurate Tempo Estimation
I'm developing a tool that makes use of madmom to generate a Syncronised beat clock to a live audio signal of a track with fixed BPM. 

The approach is to record some audio, estimate bpm and beat positions and then generate positions of all future beats. 

I have found the methods that come with madmom for tempo estimation are insufficient to accurately measure bpm. For example a track with a BPM of 124 will always be rounded up to 125 when using madmom.features.tempo.TempoEstimationProcessor. 

I have ended up hacking around the outputs to generate more accurate estimates. 
```
import madmom, scipy.stats, numpy as np
beats = madmom.features.beats.RNNBeatProcessor()("test_audio.wav")
when_beats = madmom.features.beats.BeatTrackingProcessor(fps=100)(beats)
m_res = scipy.stats.linregress(np.arange(len(when_beats)),when_beats)

first_beat = m_res.intercept 
beat_step = m_res.slope

print("bpm = ", 60/beat_step)
```

this seems to be much more accurate. Often getting the bpm to within 0.01 of the actual bpm.
Is there another way within madmom to do this sort of accurate bpm determination?

I've got a working example of how I've used this here:
https://github.com/louischaman/PyBeatSync

superbock:

This PR changes the way the tempo is computed. Formerly, only the highest peak(s) of the smoothed histogram of beat intervals was considered. Now, this histogram is interpolated (quadratic interpolation) and oversampled before determining the peak(s). This should lead to more accurate BPM values.

This pull request fixes #416.

- [ ] check if there are better ways of interpolation, e.g. interpolating only locally around the peaks
- [ ] adapt tests to reflect the higher precision

### DIFF
--- a/madmom/features/tempo.py
+++ b/madmom/features/tempo.py
@@ -193,7 +193,7 @@ def dominant_interval(histogram, smooth=None):
 
 
 # extract the tempo from a histogram
-def detect_tempo(histogram, fps):
+def detect_tempo(histogram, fps, interpolate=False):
     """
     Extract the tempo from the given histogram.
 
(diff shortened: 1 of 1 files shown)


superbock:
Ping @louischaman @hendriks73, you could try this branch to see if it produces tempo estimates closer to what you're expecting.

hendriks73:
Sounds promising! I'll take a look as soon as I can... Am a little swamped right now.

superbock:
Tempo interpolation was added in #495.

PR #428, opened 2019-04-05
fix default value for Stream.num_channels

Laubeee:
fix default value for `Stream.num_channels`
Fix: When initializing `Stream` and not providing the parameter `num_channels` there was a line that is supposed to set it to 1, but didn't.


superbock:
Thanks for the patch, merged.

PR #437, opened 2019-07-04
add key_signatures property to MidiFile, analogous to time_signatures

mgrachten:

Add a key_signatures property to MidiFile, analogous to time_signatures. It returns a structured array, because the key attribute is a string.


superbock (approved):
Thanks for this useful addition!

PR #444, opened 2019-07-29
#443: TempoDetector batch processing hangs on bad input

### LINKED ISSUE #443: TempoDetector batch processing hangs on bad input
### Expected behaviour

Run

    tempodetector --mirex batch 105247.mp3

It should finish processing, no matter what.
The attached test file is from [FMA_medium](https://github.com/mdeff/fma).

### Actual behaviour

It prints a message that the file could not be read (which is correct), but then simply hangs.

I'll provide a fix in a PR shortly.

[105247.mp3.zip](https://github.com/CPJKU/madmom/files/3440586/105247.mp3.zip)

hendriks73:
TempoDetector batch processing hangs waiting for task to finish indefinitely, if it encounters a bad input file.

Solution: Wrong Exception was caught (changed).
Made sure `task_done()` is called in a finally block.

This pull request fixes #443.

### DIFF
--- a/madmom/processors.py
+++ b/madmom/processors.py
@@ -22,6 +22,7 @@ import os
 import sys
 from collections import MutableSequence
 
+import madmom
 import numpy as np
 
 from .utils import integer_types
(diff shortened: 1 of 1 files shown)


superbock:
Thanks for spotting this error!

PR #448, opened 2019-08-20
[WIP] fix handling of non-wave files in single mode

### LINKED ISSUE #447: single mode fails with non-wave files
### Expected behaviour

Any file should be processed correctly, no matter if in `batch` or `single` mode.

### Actual behaviour

https://github.com/CPJKU/madmom/blob/master/madmom/io/audio.py#L135 overwrites the provided `infile` with `-`, thus processing of these files fails silently.

The regression was introduced with commit 33db19b572ac9df12df2a1c3af4a2441cb4312fb

Use a `.m4a` file in single mode and it will fail to process it.

superbock:

So far, only a test for reproducing issue #447 is added.

This pull request fixes #447.

- [ ] wait for response from @jpauwels to add a test for #418
- [ ] provide a fix which works in all cases


jpauwels:
Superseded by #456. The requested test was actually there all along (`TestLoadAudioFromFileObject` in `test_io_audio.py` which I added as part of #418.

PR #456, opened 2019-12-13
Fix mp4 no faststart

### LINKED ISSUE #447: single mode fails with non-wave files

jpauwels:

Fixes handling of MP4 container that are not streamable.

This pull request fixes #447 and incorporates the test of #448.


jpauwels:
Hmm, it turns out that the tests are failing because the output of SuperFlux is different on the M4A, but both for single mode (audio passed as file objects) and batch mode (audio passed as path strings), so I doubt it has anything to do with my changes. Probably a different aac decoder in ffmpeg. On my local Mac (10.14.6) the test passes with ffmpeg v.4.2.1.

superbock:
Regarding the output of SuperFlux I am definitely willing to set the absolute tolerance quite high as long as it reads those files. A cleaner solution however would be to simply use ffmpeg 4, but the most popular PPA has just disappeared...

superbock:
Could you please try rebasing your branch on the `ffmpeg_4` branch (or cherry-pick commit 32249313e0cbf8bb9da5c77606e16c9c1bba494f), which updates ffmpeg to the most current version. Unfortunately, changing the base of the PR does not change the base as git does...

jpauwels:
It looks like that solved it!

jpauwels:
I reorganised my code so that the check for non-streamable MP4 containers does not happen every time, but only when the decoding returns a byte length of zero (sadly non-streamable containers do not cause an error, but just return nothing)

madmom/io/audio.py
```diff
@@ -315,16 +312,56 @@ def decode_to_pipe(infile, fmt='f32le', sample_rate=None, num_channels=1,
...
+    if container_format == "mov,mp4,m4a,3gp,3g2,mj2" and partial_file:
+        return True
+    else:
+        return False
+
+
 def decode_to_memory(infile, fmt='f32le', sample_rate=None, num_channels=1,
-                     channel=None, skip=None, max_len=None, cmd='ffmpeg',
+                     channel=None, skip=None, max_len=None,
+                     cmd_decode='ffmpeg', cmd_probe='ffmpeg',
```
superbock: Is `ffmpeg` the correct default for `cmd_probe`? The docstring suggests that `ffprobe` should be used.

Issue #469, opened 2021-02-26
Pip packages for Python 3.8 and 3.9 missing


superbock:
### Expected behaviour

Installing via `pip install madmom` should work for newer Python versions as well.

### Actual behaviour

```
$ pip install madom
ERROR: Could not find a version that satisfies the requirement madom
ERROR: No matching distribution found for madom
```

snownstone:
Ubuntu 20.0, Python 3.8.7, I can successfully install madmom 0.16.1 by ```pip install madmom```. 

However, I met import error after I successfully installed madmom 0.17.dev0 (to use the _hpss_ submodule in audio) by ```pip install -e git+https://github.com/CPJKU/madmom#egg=madmom ```

![Selection_011](https://user-images.githubusercontent.com/61407739/111729420-ff5bf800-88a9-11eb-994c-87c2b438bd7a.png)

![Selection_010](https://user-images.githubusercontent.com/61407739/111728592-231e3e80-88a8-11eb-9f24-1c891269b2da.png)

Reinstall did not solve it.

superbock:
Installation via `pip install -e git+https://github.com/CPJKU/madmom#egg=madmom` is not supported, since the package contains code which needs to be compiled.

I will update the PyPI packages soon and include compatibility for newer Python versions. In the mean time you can install from source as described here: https://madmom.readthedocs.io/en/latest/installation.html#install-from-source

snownstone:
I read the instruction of install from source, but I did not follow it because of my confusion about the procedure in pyenv. I will try again as the suggested way.

Thank you very much for your excellent work and generous sharing.

jpauwels:
@snownstone The solution is either to update your Numpy to v1.20 or wait for my PR https://github.com/CPJKU/madmom/pull/474 to be merged.

Btw @superbock, installation via `pip install -e git+https://github.com/CPJKU/madmom#egg=madmom` should work (again) after that, it's not the cause of this error. The reason why `pip install madmom` works in this case is that it uses a wheel that has been build before Numpy v1.20 was released. If you'd build wheels for madmom v0.16.1 and/or Python 3.8 or 3.9 at this moment, you'd risk running into the same problem.

danuo:
> @snownstone The solution is either to update your Numpy to v1.20 or wait for my PR #474 to be merged.
> 
[...]

PR #474, opened 2021-05-06
Pin numpy version at build to handle Numpy's API change in v1.20

jpauwels:

The [C-API change in Numpy v1.20](https://numpy.org/doc/stable/release/1.20.0-notes.html#c-api-changes) has the potential to cause incompatibility issues, as seen in https://github.com/CPJKU/madmom/issues/469#issuecomment-802535058.

This happens when a user hasn't upgraded yet to Numpy 1.20 (or can't upgrade, e.g. because of apt-get availability or pinned due to Tensorflow requirements). Because the build requirement in `pyproject.toml` is an unversioned numpy, it will pull in the latest >=1.20 version and build madmom against it, which gives problems when running with the <1.20 version because the latter is not forward compatible.

Numpy's C-API is forward compatible though, so the solution is to pin the build version at a low enough version, it doesn't matter if this version is lower than the version used at runtime. You could do this per Python version, as is currently done in `pyproject.toml`, but there even is a meta-package for this since it's a common problem. Using the meta-package also makes the solution robust against future Python versions, so that's what I did for this PR.

I also snuck in an unrelated warning fix that will become an error in Python 3.10 (due in October).

### DIFF
--- a/madmom/processors.py
+++ b/madmom/processors.py
@@ -20,7 +20,7 @@ import itertools as it
 import multiprocessing as mp
 import os
 import sys
-from collections import MutableSequence
+from collections.abc import MutableSequence
 
 import madmom
 import numpy as np
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,6 @@ requires = [
     "setuptools",
     "wheel",
     "cython>=0.25",
-    "numpy>=1.13.4;python_version!='3.5'",
-    "numpy>=1.13.4,<=1.15.2;python_version=='3.5'"
+    "oldest-supported-numpy"
 ]
 build-backend = "setuptools.build_meta"


superbock:
Thanks for the fix! I was not aware of this rather elegant fix :)

mohkoma:
Is there any update yet? as you said Tensorflow require < 1.20 version.
Thanks!

faroit:
@superbock this issue is blocking a JOSS review in progress https://github.com/openjournals/joss-reviews/issues/3391 and prevents one of the reviewers (@keunwoochoi) to evaluate the performance of the software. It would be great if a new madmom release can be put out soon! Thanks for your great work

keunwoochoi:
Thank you all!

faroit:
@superbock could you issue a new madmom pypi release with this change included to help dependencies resolving? 🙏

samuelbradshaw:
Hi! It looks like this was merged a couple of years ago, but there hasn't been a release since. Is there a madmom release planned soon?

Issue #489, opened 2021-11-30
No-beat log density of RNNDownBeatTrackingObservationModel can be invalid


omareldeeb:
### Expected behaviour

Log densities of the observations should be valid, non-inf, and non-nan floating-point numbers.

### Actual behaviour

If `np.sum(observations, axis=1)` is greater than 1, the log of a negative number is attempted to be calculated in line 629 of `beats_hmm.py`, causing a `nan` to be stored in `log_densities`. This behavior causes the viterbi algorithm to potentially fail.  

### Steps needed to reproduce the behaviour

```python
import madmom
import numpy as np
import scipy

beat_decoder = madmom.features.downbeats.DBNDownBeatTrackingProcessor(fps=100.0,
                                                                      correct=True,
                                                                      beats_per_bar=4)
beat_activation = np.random.uniform(0, 1, 100)
downbeat_activation = np.random.uniform(0, 1, 100)
observations = np.dstack((beat_activation, downbeat_activation))[0]

print(beat_decoder.hmms[0].observation_model.log_densities(observations))
```

### Information about installed software

madmom version: '0.16.1'
numpy version: '1.19.5'
scipy version: '1.7.2'

superbock:
The observations (including the observations for non-beat states) have to be  a probability density function in order to work with `DBNDownBeatTrackingProcessor`. Thus your observations are not suitable for `DBNDownBeatTrackingProcessor`. I will update the documentation to be more clear about that.

PR #490, opened 2021-11-30
Fix bad log densities

### LINKED ISSUE #488: IndexError in self.correct branch in DBNDownBeatTrackingProcessor process function.
### Expected behaviour

An empty array to be returned, if no valid path could be decoded using the Viterbi algorithm.

### Actual behaviour

If `self.correct == True`, an IndexError is thrown in line 301 of `downbeats.py` (), due to `beat_range` being empty.

```python
import madmom
import numpy as np
import scipy

beat_tracker = madmom.features.downbeats.DBNDownBeatTrackingProcessor(fps=100.0,
                                                                                                                                        correct=True,
                                                                                                                                        beats_per_bar=4)
beat_activation = np.random.uniform(0, 1, 100)
downbeat_activation = np.random.uniform(0, 1, 100)

beat_decoder.process(np.dstack((beat_activation, downbeat_activation))[0])

```

### Information about installed software

Please provide some information about installed software.

madmom.__version__ == '0.16.1'
np.__version__ == '1.19.5'
scipy.__version__ == '1.7.2'

omareldeeb:

Please provide a detailed description of the changes proposed in this pull
request. This can either be a textual description or a simple list.

In line 629 of `beats_hmm.py`, the sum of the beats and downbeats observation is used to calculate the log density for non-beat states. These, however, are not guaranteed to define a joint probability density function, which causes the sum to be potentially greater than 1 and the logarithm of a negative number to be used. Since beats are also downbeats, we can simply take the beat observation instead.

If the viterbi algorithm fails to decode a valid path (e.g. as a result of the above error), it returns an empty array, which is indexed in line 305 of `downbeats.py`, causing an IndexError. To avoid this, we can simply check if the array is empty in line 297.

This pull request fixes #488 #489 .

### DIFF
--- a/madmom/features/beats_hmm.py
+++ b/madmom/features/beats_hmm.py
@@ -628,7 +628,7 @@ class RNNDownBeatTrackingObservationModel(ObservationModel):
         log_densities = np.empty((len(observations), 3), dtype=np.float)
         # Note: it's faster to call np.log multiple times instead of once on
         #       the whole 2d array
-        log_densities[:, 0] = np.log((1. - np.sum(observations, axis=1)) /
+        log_densities[:, 0] = np.log((1. - observations[:, 0]) /
                                      (self.observation_lambda - 1))
         log_densities[:, 1] = np.log(observations[:, 0])
         log_densities[:, 2] = np.log(observations[:, 1])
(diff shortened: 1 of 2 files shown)


superbock:
This is only partly correct. The output of `RNNDownBeatProcessor` is actually a probability density function. Thus the situation outlined in #489 should actually not happen — at least not with `RNNDownBeatProcessor`.

The proposed fix changes the behaviour of `DBNDownBeatTrackingProcessor` since it actually expects the input to be a probability density function (or at least look like one), i.e. it is designed to have lower beat activations at downbeat positions.

PR #505, opened 2022-06-18
Fixing multiprocessing memory leak in downbeats.py

igalcohenhadria:
Moving the multiprocessing pool initialisation to the process area.
(using multiprocessing.pool in the init section cause the pool to never release thus creating zombies)

Please provide a detailed description of the changes proposed in this pull
request. This can either be a textual description or a simple list.

-
-

This pull request fixes #.

Please remove this section (inluding the headline above) if there aren't any
open TODOs. If there are, please remove this text, fill the items below and
add a [WIP] prefix to the pull request's name.

- [ ] TODO item description

### DIFF
--- a/madmom/features/downbeats.py
+++ b/madmom/features/downbeats.py
@@ -227,12 +227,7 @@ class DBNDownBeatTrackingProcessor(Processor):
                              'and `transition_lambda` must all have the same '
                              'length.')
         # get num_threads from kwargs
-        num_threads = min(len(beats_per_bar), kwargs.get('num_threads', 1))
-        # init a pool of workers (if needed)
-        self.map = map
-        if num_threads != 1:
-            import multiprocessing as mp
-            self.map = mp.Pool(num_threads).map
+        self.num_threads = min(len(beats_per_bar), kwargs.get('num_threads', 1))
         # convert timing information to construct a beat state space
         min_interval = 60. * fps / max_bpm
         max_interval = 60. * fps / min_bpm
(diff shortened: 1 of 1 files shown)


superbock:
That's a bit weird, since I exactly did the opposite — moving the pool creation to `__init__()` — because it was leaking memory and file descriptors when being set up in `process()`, [at least this is what the comment I left in `processors.py` suggests](https://github.com/CPJKU/madmom/blob/3bc8334099feb310acfce884ebdb76a28e01670d/madmom/processors.py#L448).

May I ask in which situations these zombie processes appear?

If there's a problem with zombies, we should probably aim for a dedicated process/pool destruction mechanism or use a dedicated pooling solution which does what we need.

igalcohenhadria:
Hello,

I was using the processor in a simple loop and ended up with a memory allocation failure after a few thousands … 

I tracked down the problem there and fixed it for me the way I sent in the pool request … 

I ll try to send you more info tomorrow if you want 

Envoyé de mon iPhone

superbock:
Maybe you can paste a snippet of the code you're using. Using the processor in a loop doesn't sound too efficient... usually it's much better to instantiate the processor only once and then call it either in a loop or with all files at once and let it do the processing (in parallel) for you.

igalcohenhadria:
Here the code and the running test in a youtube video.

https://www.youtube.com/watch?v=4DY8wW19QrE

The code just loop freely and you will see zombies appearing in the processor counter i made.
Initializing once may be effective to limit the number of zombies, but after python3 stops, zombies would still remains.

Also good practice around multiprocessing is to use the pool inside a "with" or closing the pool at some point manually.

An alternative to the patch i gave would be to keep the pool as an attribute in the __init__ and close the pool in the __del__())

PR #540, opened 2024-08-23
CI and NumPy compatibility updates

jpauwels:
Hi @superbock, looks like you could use some superpauwels ;-)

Would hate to see `madmom` bitrotting away, so I did some maintenance jobs. I fixed the CI, then made some changes to fix compatibility with NumPy 2. I took the liberty of removing Python 3.8 from the list of supported versions (end of life in October, so not worth the trouble imho) and added Python3.11 and 3.12. All checks pass on [my CI run](https://github.com/jpauwels/madmom/actions/runs/10525029422).

Hope this helps to get a version 0.17 released quickly!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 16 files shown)


superbock:
Wow, thanks a lot! I'll have a deeper look at it hopefully tomorrow.

Issue #527, opened 2023-08-24
the beat tracking rely on numpy=<1.19


nicolaus625:
from madmom.features.beats import DBNBeatTrackingProcessor
  File "/.conda/envs/muvi/lib/python3.9/site-packages/madmom/__init__.py", line 24, in <module>
    from . import audio, evaluation, features, io, ml, models, processors, utils
  File "/.conda/envs/muvi/lib/python3.9/site-packages/madmom/evaluation/__init__.py", line 874, in <module>
    from . import chords, beats, notes, onsets, tempo
  File "/.conda/envs/muvi/lib/python3.9/site-packages/madmom/evaluation/chords.py", line 35, in <module>
    from ..io import load_chords
  File "/.conda/envs/muvi/lib/python3.9/site-packages/madmom/io/__init__.py", line 22, in <module>
    SEGMENT_DTYPE = [('start', np.float), ('end', np.float), ('label', object)]
  File "/.conda/envs/muvi/lib/python3.9/site-packages/numpy/__init__.py", line 305, in __getattr__
    raise AttributeError(__former_attrs__[attr])
AttributeError: module 'numpy' has no attribute 'float'.
`np.float` was a deprecated alias for the builtin `float`. To avoid this error in existing code, use `float` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.float64` here.
The aliases was originally deprecated in NumPy 1.20; for more details and guidance see the original release note at:
    https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations

The error message you're encountering indicates that there's a deprecated usage of np.float in the Madmom library's io module. The message suggests that np.float is no longer a valid attribute in recent versions of NumPy (starting from version 1.20), and you should use either the built-in float type or np.float64 if you specifically need a NumPy scalar type.

To resolve this issue, you need to update the affected code in the Madmom library. Specifically, you should replace instances of np.float with either float or np.float64, depending on the context.

For example, if the problematic line of code is:

```SEGMENT_DTYPE = [('start', np.float), ('end', np.float), ('label', object)]```
You can modify it to use float:

```SEGMENT_DTYPE = [('start', float), ('end', float), ('label', object)]```
Or, if you want to use the NumPy scalar type:

```SEGMENT_DTYPE = [('start', np.float64), ('end', np.float64), ('label', object)]```
[...]

PR #542, opened 2024-12-10
Replace deprecated NumPy types (np.int and np.float) with int and float

JinruG:

NumPy deprecated np.int and np.float starting from version 1.20. Using these deprecated types triggers warnings or errors in modern Python environments. To ensure compatibility with newer NumPy versions, the code has been updated to use the recommended replacements.

This pull request fixes #.

Replaced all instances of np.int with int.
Replaced all instances of np.float with float.
Verified that functionality and type expectations remain consistent across all affected functions and classes.
These changes were tested by running the existing test suite using Python 3.9.7 and NumPy 1.24. No regressions or warnings were observed.

This Pull Request addresses https://github.com/CPJKU/madmom/issues/527 

@superbock


jvaleroliet:
Please!

